PR #85, opened 2014-11-10
Allow exception handler on command line

bradleyy:
Allow the user to pass in an exception handler on the command line.


selwin:
The implementation of passing in exception handler should mimic [RQ's](https://github.com/nvie/rq/blob/master/rq/cli/cli.py#L145) in that supports multiple exception handlers

PR #89, opened 2014-12-04
fix(django-rq/templates): updated job.status to job.get_status

agconti:
When using the django_rq.urls, a depreciation warning is thrown because job.status is depreciated in python-rq. This prevents you from viewing the jobs entirely. The templates have been updated accordingly.

Closes #88.

PR #90, opened 2014-12-10
Fix job.get_status check on job detail page.

### LINKED ISSUE #88: Viewing job queues fails because `job.status` is depreciated.
When trying to look at a job's details using the provided `django-rq.urls`, a `DeprecationWarning` is thrown when trying to render the template.

```python
DeprecationWarning at /django-rq/queues/0/28708d4d-57e6-4a31-b40e-b91f50510c99/
job.status is deprecated. Use job.get_status() instead 
```

It seems that `rq` has deprecated `job.status` and the templates need to be updated accordingly.

zaius:
Should have caught this in #87 but I missed the get_status call in the job_detail page. Fixes #88. Provides better backwards compatibility than #89.

### DIFF
--- a/django_rq/templates/django_rq/job_detail.html
+++ b/django_rq/templates/django_rq/job_detail.html
@@ -79,7 +79,7 @@
             <div>
                 <label class="required">Status:</label>
                 <div class="data">
-                    {% if job.get_status %}
+                    {% if job.get_status|length != 0 %}
                         {{ job.get_status }}
                     {% else %}
                         {{ job.status }}


selwin:
The master branch of django_rq requires RQ >= 0.5 so we no longer needs this. Thanks for the PR!

zaius:
I think that's a much nicer solution :-)

The if statement is redundant in both the jobs.html and job_detail.html pages now. I've already fixed that on my local copy - would you like a new PR to fix them?

selwin:
Yes please :)

PR #92, opened 2015-02-06
Allow connection to use StrictRedis instead of Redis

wastrachan:
The addition of a `strict_redis` kwarg on `get_connection()` allows projects that want (or need) to take advantage of `redis.StrictRedis()` to do so in a more pythonic way: without having to manually set up said connections.

### DIFF
--- a/django_rq/queues.py
+++ b/django_rq/queues.py
@@ -48,8 +48,10 @@ def get_redis_connection(config, use_strict_redis=False):
     """
     Returns a redis connection from a connection config
     """
+    redis_cls = redis.StrictRedis if use_strict_redis else redis.Redis
+
     if 'URL' in config:
-        return redis.from_url(config['URL'], db=config['DB'])
+        return redis_cls.from_url(config['URL'], db=config['DB'])
     if 'USE_REDIS_CACHE' in config.keys():
 
         from django.core.cache import get_cache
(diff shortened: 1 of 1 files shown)


selwin:
Thanks for the PR. Do you mind changing the kwarg to "use_strict_redis"?

wastrachan:
Good suggestion, I agree.

wastrachan:
That's a good suggestion.

django_rq/queues.py
```diff
@@ -72,21 +72,29 @@ def get_redis_connection(config):
         else:
             # We're using django-redis-cache
             return cache._client
-        
+
```
selwin: We can make this a little bit cleaner by doing this:
```python
redis_class = redis.StrictRedis if use_strict_redis else redis.Redis

if 'UNIX_SOCKET_PATH' in config:
    return redis_class(unix_socket_path=config['UNIX_SOCKET_PATH'], db=config['DB'])

return redis_class(host=config['HOST'],
                   port=config['PORT'], db=config['DB'],
                   password=config.get('PASSWORD', None))
```
Mind making these changes? I'll merge this in right after

wastrachan: You know, I thought about this approach when I originally created this PR. I originally didn't do it this way because I (incorrectly) assumed that `StrictRedis` did not have a `from_url` method. It appears I was wrong-- I'll make this change, it's much cleaner.

selwin:
Thanks for the PR!

PR #104, opened 2015-04-09
Update django-rq compatibility with django 1.7/1.8

SeiryuZ:
Hi,

There's still warning on test with django 1.8
```
RemovedInDjango19Warning: Loading the `url` tag from the `future` library is deprecated and will be removed in Django 1.9. Use the default `url` tag instead.
  RemovedInDjango19Warning)
```

As for the test ``test_empty_queue_setting_raises_exception``, I deleted it since I can't figure out how to fix it.

Before django 1.7, the `django_rq.settings` file is loaded AFTER the test function is called.

However in 1.7 and 1.8, the `django_rq.settings` file is loaded BEFORE the test function, hence the QUEUES variable will always have a value and results in no exception.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,14 +6,20 @@ python:
   - "3.3"
 
 env:
-  - DJANGO=1.6.1
-  - DJANGO=1.5.1
-  - DJANGO=1.4.5
+  - DJANGO=1.8.0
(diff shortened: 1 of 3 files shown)


selwin:
Thanks for the PR!

PR #118, opened 2015-07-30
Sort the queues by name

Microserf:
This change makes the ordering of queues deterministic across multiple instances of `django-rq`. This means that in load-balanced environments (like Heroku), a path like `/rq/queues/1` will point to the same queue regardless of the instance that serves the request.


selwin:
Thanks for the PR!

Issue #120, opened 2015-08-14
Include rq-scheduler jobs in dashboard


marksteve:

taewookim:
It's been 5 years... any updates?

selwin:
This project already shows scheduled jobs from RQ's `ScheduledJobRegistry`.

PR #121, opened 2015-08-14
Add view for scheduled jobs

marksteve:
Attempt on #120

This will probably break with a huge list of scheduled jobs since there's no pagination.
Not sure how to approach that properly though.


django_rq/views.py
```diff
@@ -51,6 +52,11 @@ def stats(request):
             queue_data['finished_jobs'] = len(finished_job_registry)
             queue_data['started_jobs'] = len(started_job_registry)
             queue_data['deferred_jobs'] = len(deferred_job_registry)
+            try:
+                scheduler = get_scheduler(queue.name)
+                queue_data['scheduled_jobs'] = len(scheduler.get_jobs())
```
selwin: This line could potentially be very inefficient as it fetches all scheduled jobs. I'd rather we implement a `scheduler.count()` first before we merge this in.

marksteve: Yeah. I'll try to do a PR on it in the scheduler repo.

selwin: Thanks!

smaccona: I started to implement a `scheduler.count()` method using Redis `ZCOUNT`, but if you have multiple RQ queues defined in Django (in the same Redis instance), `ZCOUNT` will just return all jobs scheduled for all queues, as `scheduler.get_jobs()` does now when it retrieves all jobs.

The `origin` of the `Job` has the name of the queue the job will be eventually queued on, but I can't think how to filter for that at the Redis level without retrieving all jobs. I can't think of an efficient way to get the count of jobs scheduled for each queue without retrieving all jobs and manually filtering.

I have made an issue at https://github.com/ui/rq-scheduler/issues/118 and will make the PR for `count()` anyway, perhaps with an optional queue name.

selwin:
@marksteve thanks for the PR. Before this could go in though, there's a few things that need to change:

1. https://github.com/ui/rq-scheduler/issues/122 needs to be implemented so we could page through scheduled jobs without DOS'ing servers with large number of scheduled jobs :)
2. We should not provide a way to view scheduled jobs on a per queue basis but rather as a scheduler for all queues (please see how views for `FailedQueue` is implemented)

Once again thanks for working on this and sorry for the late reply.

selwin:
I have released rq-scheduler 0.6.1 which has `scheduler.count` method. Please update this PR when you have the time @marksteve :)

marksteve:
See https://github.com/ui/rq-scheduler/pull/119

smaccona:
@marksteve since you closed this, I am going to open a separate PR for the view - the two PRs I submitted just provided the underlying required support in `rq-scheduler`, a separate PR for the view in `django-rq` is still required.

tomstoreysn:
Any movement on this? This would be so so handy to have!

smaccona:
Hi @tomstoreysn - I dropped the ball on this since it wasn't a quick fix. It would be very handy, though, so I plan to open a PR for this in the next day or two.

@selwin the `Scheduler` class in [RQ Scheduler](https://github.com/ui/rq-scheduler) doesn't implement several of the functions we need for the DjangoRQ views (e.g. `name` property, `count` property, `key` property, `empty()` function and `get_jobs()` function), so I will implement them in a proxy class within DjangoRQ. RQ's `FailedQueue` does implement those methods, so they weren't needed to implement the `failed` queue views. It's not straightforward to implement these directly in the `Scheduler` class because there are at least two conflicts:

1. The `count()` function is decorated with `@property` in RQ's `Queue` class, but is just a plain function in the `Scheduler` class. I don't want to change this as it could break existing code.
2. The `get_jobs` function in `Scheduler` has a different signature that in RQ's `Queue`, and we also want to capture the scheduled time for each job which is not available in the `Job` itself by default.

I'll submit a PR in the next day or two.

selwin:
@smaccona I don't like having a proxy class because it just adds another layer of indirection in interacting with the scheduler.

For the initial page, I think we should add a new scheduler section here (instead of mixing schedulers with queues).

selection_418

As for the page listing jobs in the scheduler, it would be better to make a separate view function instead of reusing view used to display list of jobs within a queue. This also gives us the flexibility to show different information like scheduled enqueue time for each job.

smaccona:
@selwin fair enough. I can remove the proxy class and incorporate most of that functionality directly into the views.

I am taking your suggested approach in treating the `scheduled` queue like `failed`, but I left `failed` in the last position (I suspect people are used to seeing it there):

image

For the functionality that I had in the proxy class, I think it makes sense to move the `empty()` method over to [RQ Scheduler](https://github.com/ui/rq-scheduler), so I will submit a PR there for that.

Once that is in place, the changes to the existing view methods are pretty minor, even the changes to the templates. If I create completely separate view methods and templates, there will be a lot of duplication. Can you look at the diffs at https://github.com/smaccona/django-rq/commit/afc4180ea8439f9dcfdb142f93295820d12bafbf and see if you are happy with my current approach? If not, I can create separate view methods and templates.

selwin:
It would be better to put scheduler in a separate section, I think. Code duplication are also fine for me if it makes sense (and in this case the `Scheduler` is really a different object than a `Queue`).

I plan to replace `FailedQueue` with `FailedJobRegistry` (specific per queue) in the next few releases because it doesn't scale as well as the new registry based implementation.

Thanks a lot for your effort @smaccona !

smaccona:
@selwin - haven't forgotten about this, just tied up with work. I will come up with an approach more in line with your overall development direction.

eriktelepovsky:
Hi. Any update for this great upcoming functionality, please?

taewookim:
Ditto @eriktelepovsky 
Any updates?

PR #126, opened 2015-10-05
allow to specify db number in url

anentropic:
when doing config via env vars it's nice to be able to pass the whole configuration in one url string

underlying redis-py library supports this


selwin:
Thanks for the PR!

PR #128, opened 2015-10-09
Add message about django-suit integration

rsenk330:
We added some templates to make the admin look nice with [django-suit](http://djangosuit.com/). I'm not sure if this is something you'd be interesting in including or not, so I added README. If you do want to include it, but have a better place, let me know and I can move it.


selwin:
More documentation is always good :). Thanks for the PR!

PR #130, opened 2015-10-23
Compatibility with Django 1.9

douglas:
I was testing Django 1.9 and it needs the following fixes:

* Remove 1.9 deprecation warnings in urls.py
* Removing {% load url from future %} because it is deprecated

Thanks for **django-rq**, it is awesome =)

PS: Perhaps, you should merge in a django_1.9 branch !

### DIFF
--- a/django_rq/urls.py
+++ b/django_rq/urls.py
@@ -1,24 +1,22 @@
 from django.conf.urls import url
-from .views import stats, jobs, finished_jobs, started_jobs, deferred_jobs
-from .views import clear_queue, job_detail, delete_job, actions, requeue_job_view
-
+import views
 
 urlpatterns = [
-    url(r'^$', stats, name='rq_home'),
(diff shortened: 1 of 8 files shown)


django_rq/urls.py
```diff
@@ -1,21 +1,24 @@
-from django.conf.urls import patterns, url
```
selwin: Could you please revert this file back to how it was? I prefer the old approach as it has less imports :)

aidanlister: The old approach is deprecated (and ugly!), see https://docs.djangoproject.com/en/1.8/ref/urls/#patterns

django_rq/urls.py
```diff
@@ -1,21 +1,22 @@
...
+from django.conf.urls import url
+import views
```
aidanlister: this should be "from . import views" (otherwise it won't work in py3)

shulcsm:
+1

PR #131, opened 2015-11-02
Add 'worker-ttl' option to rqworker command

pnuckowski:


selwin:
Sorry, I couldn't merge this PR in because of a conflict, but I made a different commit: https://github.com/ui/django-rq/commit/75f266d6d85ca864bf24e57b0bd737c239c8db71

PR #135, opened 2015-12-02
Removed 'load url from future' for compatibility with Django 1.9

viaregio:
From the docs of the django project:

One deprecated feature worth noting is the shift to “new-style” url tag. Prior to Django 1.3, syntax like {% url myview %} was interpreted incorrectly (Django considered "myview" to be a literal name of a view, not a template variable named myview). Django 1.3 and above introduced the {% load url from future %} syntax to bring in the corrected behavior where myview was seen as a variable.

The upshot of this is that if you are not using {% load url from future %} in your templates, you’ll need to change tags like {% url myview %} to {% url "myview" %}. If you were using {% load url from future %} you can simply remove that line under Django 1.5

This tag shouldn't be necessary anymore since Django 1.5.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,22 +4,25 @@ python:
   - "2.6"
   - "2.7"
   - "3.3"
+  - "3.4"
 
 env:
-  - DJANGO=1.8.0
+  - DJANGO=1.9.0
+  - DJANGO=1.8.7
   - DJANGO=1.7.7
   - DJANGO=1.6.11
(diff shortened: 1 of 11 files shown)


selwin:
Thanks for the PR! Could you please remove Django 1.4 from travis.yml and add Django 1.9? We should also indicate that this project requires Django 1.5+ to use in Readme and setup.py.

selwin:
Thanks for the PR!

PR #138, opened 2015-12-07
Use future-proof URLconf format.

aaugustin:
This syntax is compatible with all past and, as far as the current
deprecation timeline is concerned, future versions of Django.


aaugustin:
This avoids a bunch of deprecation warnings when running on Django 1.9.

aaugustin:
Note that #138 contains pretty much the same changes, showing some demand for this fix and a release on PyPI.

selwin:
Merged, thanks for the delay!

PR #149, opened 2016-01-12
Added pid option that mimics the same option on RQ's cli

ydaniv:
This is practically a copy-paste from RQ's cli (:


selwin:
Thanks for the PR!

ydaniv:
Thanks! :+1:

PR #151, opened 2016-01-18
makes queue name configurable for rqscheduler

sbussetti:


selwin:
Rq-scheduler is not built around the concept of having multiple schedulers running at the same time so I don't think this is a very good idea. Thanks for the PR though.

sbussetti:
Sure you could do that, but without the ability to configure the queue name, you can't run your scheduler on a separate dedicated queue.

PR #153, opened 2016-01-31
Django 1.9 compatibility

### LINKED ISSUE #142: jquery.js not found in the admin jobs list page in Django 1.9
This issue is about
https://github.com/ui/django-rq/blob/master/django_rq/templates/django_rq/jobs.html#L12
I see 404 Error near the jquery.js url, and JS doesn't works.

In the Django 1.9:

> The jQuery static files in contrib.admin have been moved into a vendor/jquery subdirectory

https://docs.djangoproject.com/en/1.9/releases/1.9/#miscellaneous

Technically, it's located at admin/js/vendor/jquery/jquery.js now.

philippbosch:
Fixes #142

### DIFF
--- a/django_rq/templates/django_rq/clear_queue.html
+++ b/django_rq/templates/django_rq/clear_queue.html
@@ -1,5 +1,7 @@
 {% extends "admin/base_site.html" %}
 
+{% load admin_static %}
+
 {% block extrastyle %}
     {{ block.super }}
     <style>
@@ -9,7 +11,7 @@
             width: 80%;
         }
     </style>
-    <link href="{{ STATIC_URL }}admin/css/forms.css" type="text/css" rel="stylesheet">
+    <link href="{% static 'admin/css/forms.css' %}" type="text/css" rel="stylesheet">
 {% endblock %}
 
 {% block breadcrumbs %}
--- a/django_rq/views.py
+++ b/django_rq/views.py
@@ -1,13 +1,11 @@
 from __future__ import division
 
-from distutils.version import StrictVersion
 from math import ceil
 
 from django.contrib import messages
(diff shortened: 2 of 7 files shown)


django_rq/views.py
```diff
@@ -84,6 +92,7 @@ def jobs(request, queue_index):
         'page': page,
         'page_range': page_range,
         'job_status': 'Queued',
+        'jquery_path': JQUERY_PATH,
```
selwin: Instead of sprinkling jquery_path variable everywhere, can we just write a simple template tag that outputs the right path?

django_rq/templates/django_rq/clear_queue.html
```diff
@@ -9,7 +11,7 @@
             width: 80%;
         }
     </style>
-    <link href="{{ STATIC_URL }}admin/css/forms.css" type="text/css" rel="stylesheet">
+    <link href="{% static 'admin/css/forms.css' %}" type="text/css" rel="stylesheet">
```
selwin: Does STATIC_URL no longer work in 1.9?

philippbosch: It didn't work for me, and since you've been using the `{% static %}` tag everywhere else I thought it might be a good idea to make it consistent.

django_rq/views.py
philippbosch: good idea, did that in ec0e894. what do you think?

philippbosch:
@selwin your proposed changes have been incorporated. what do you think?

selwin:
Sorry for the delay and thanks for working on this PR :)

philippbosch:
No worries :) thanks for merging!

PR #158, opened 2016-03-02
Make *_at fields in Job class timezone aware

### LINKED ISSUE #32: Non-local timestamps in the admin interface
The django-rq's admin interface shows timestamps in UTC, instead of the local timezone. This is inconsistent with how the django admin interface works.

mjec:
Should fix #32 without causing any issue if `pytz` is not installed.

Create a new sublcass of `rq.job.Job` called `django_rq.job.TZAwareJob`. This makes fields in `job.TIMEZONE_AWARE_FIELDS` timezone aware using `django.utils.timezone.make_aware`.

We then modify `DjangoRQ` and `views` to use `TZAwareJob` instead of `Job`.

A couple of thoughts worth considering before pulling to master:
- might be worth checking django's USE_TZ setting?
- in my tests, my queue list still doesn't appear with `created_at` and `enqueued_at` timezone aware. Not sure why this is?

### DIFF
--- a/django_rq/queues.py
+++ b/django_rq/queues.py
@@ -6,6 +6,8 @@ from rq.queue import FailedQueue, Queue
 
 from django_rq import thread_queue
 
+from .job import TZAwareJob
+
 
 def get_commit_mode():
     """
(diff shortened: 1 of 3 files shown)


django_rq/queues.py
```diff
@@ -32,7 +34,9 @@ def __init__(self, *args, **kwargs):
         autocommit = kwargs.pop('autocommit', None)
         self._autocommit = get_commit_mode() if autocommit is None else autocommit
 
-        return super(DjangoRQ, self).__init__(*args, **kwargs)
+        job_class = kwargs.pop('job_class', TZAwareJob)
```
selwin: Thanks for the PR! I don't think overriding the `job_class` argument is a good idea because that will prevent people from using their own custom job class.

I think a better solution would be to create a template filter that displays the time in user's default timezone.

selwin:
This is fixed in master branch.

PR #160, opened 2016-03-10
Clean up EXCEPTION_HANDLER implementation

sbussetti:
I found there were a couple problems with the implementation for EXCEPTION_HANDLERS and cleaned them up.

1. The README documentation didn't match how it was implemented.
2. The implementation bypassed the django_rq settings and accessed an `RQ` settings variable directly.


selwin:
Thanks for the PR!

PR #162, opened 2016-03-21
Add scheduled jobs column

sburns:
This adds:
- [x] A column to display scheduled jobs per queue on the stats page when [rq-scheduler](https://github.com/ui/rq-scheduler) is installed.
- [x] A view to list scheduled jobs for the queue.
- [x] Tests

I didn't dig deep much into the `jobs.html` template though most of the columns are not useful for schedulued jobs.

### DIFF
--- a/django_rq/templates/django_rq/stats.html
+++ b/django_rq/templates/django_rq/stats.html
@@ -75,7 +75,7 @@
                     {% if display_scheduled_jobs %}
                         <td>
                             {% if queue.name != 'failed' %}
-                                <a href = "todo">
+                                <a href = "{% url 'rq_scheduled_jobs' queue.index %}">
                                     {{ queue.scheduled_jobs }}
                                 </a>
                             {% else %}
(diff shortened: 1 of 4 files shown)


sburns:
I noticed if you run the test suite twice, the second run fails even on `master`. `$ 127.0.0.1:6379 > DEL rq:deferred:django_rq_test` fixes it. Known issue?

sburns:
I think #121 is the better PR, does that need a PR adding the `scheduler.count` method?

selwin:
I'll leave the same comment here as I did on #121:

Thanks for the PR. Before this could go in though, there's a few things that need to change:

1. https://github.com/ui/rq-scheduler/issues/122 needs to be implemented so we could page through scheduled jobs without DOS'ing servers with large number of scheduled jobs :)
2. We should not provide a way to view scheduled jobs on a per queue basis but rather as a scheduler for all queues (please see how views for `FailedQueue` is implemented)

Once again thanks for working on this and sorry for the late reply.

chaffeqa:
Any movement on this? would be a great addition!

tomstoreysn:
It would be great to see this merged, any chance this can be looked at/when could it be completed?

Being able to view scheduled jobs would be incredibly useful!

quantumlink:
Still active? I may create a PR if not. 👍

selwin:
Yes. Please create a PR for this

selwin:
Closing this idle PR.

eriktelepovsky:
+1

PR #163, opened 2016-04-15
sort the queues only by name

pnuckowski:
It sorts RQ_QUEUES only by key (queue name). 
This issue was already resolved in #118  but unfortunately it doesn't work in our environment (behind load-balancer). 
`.../django-rq/queues/3/` each request point to the different queue.


selwin:
Thanks!

PR #168, opened 2016-06-05
replaced options with ArgumentParser

jtburchfield:
fixed django 1.10 warning.


lechup:
I guess we need to update travis matrix to oficially supported django versions:
https://github.com/ui/django-rq/pull/173

PR #171, opened 2016-06-23
Custom queue class.

Krukov:
I don't know why It is not implemented yet, but I think it is useful. For example: We have the project,  and we don't want to  run redis at the test environment or in other special environment. Yeah, writing something like this: 
```python
if getattr(settings, 'DIRECT_CALL', False):
    job()
else:
   job.delay()
```
at every call is not good idea (for me)
Next changes give for us more clear way to do this, I can use custom Queue class for this:
```python
class DummyQueue(DjangoRQ):
    def enqueue_call(self, func, args=None, kwargs=None, timeout=None,
                     result_ttl=None, ttl=None, description=None,
                     depends_on=None, job_id=None, at_front=False):
        return partial(func, args=args, kwargs=kwargs)
``` 
and then use them at the settings.
I think somebody will find other cases to use this ability: adding extra logic for new jobs, or adding special logs, etc.

### DIFF
--- a/django_rq/queues.py
+++ b/django_rq/queues.py
@@ -34,7 +34,13 @@ def get_queue_class(config):
     elif 'QUEUE_CLASS' in RQ:
         queue_class = RQ.get('QUEUE_CLASS')
 
-    if isinstance(queue_class, basestring):
+    try:
+        from django.utils import six
+        string_type = six.string_types
+    except ImportError:
+        string_type = basestring
+
+    if isinstance(queue_class, string_type):
         queue_class = import_attribute(queue_class)
     return queue_class
 
--- a/django_rq/queues.py
+++ b/django_rq/queues.py
@@ -23,6 +23,22 @@ def get_commit_mode():
     return RQ.get('AUTOCOMMIT', True)
 
 
+def get_queue_class(config):
+    """
+    Return queue class from config or from RQ settings, otherwise return DjangoRQ
+    """
+    RQ = getattr(settings, 'RQ', {})
+    queue_class = DjangoRQ
+    if 'QUEUE_CLASS' in config:
+        queue_class = config.get('QUEUE_CLASS')
+    elif 'QUEUE_CLASS' in RQ:
+        queue_class = RQ.get('QUEUE_CLASS')
+
+    if isinstance(queue_class, basestring):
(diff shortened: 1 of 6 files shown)


django_rq/queues.py
```diff
@@ -118,10 +141,11 @@ def get_queue(name='default', default_timeout=None, async=None,
 
     if default_timeout is None:
         default_timeout = QUEUES[name].get('DEFAULT_TIMEOUT')
-
-    return DjangoRQ(name, default_timeout=default_timeout,
-                    connection=get_connection(name), async=async,
-                    autocommit=autocommit)
+    if queue_class is None or queue_class == DjangoRQ:
```
selwin: I think this should simply be (if you specify anything at all, no need to do anything special with it).
```python
if queue_class is None:
    queue_class = get_queue_class(QUEUES[name])
```

selwin:
Besides my minor comment. This PR looks good. Could you please document this feature in the management command and options sections? Thanks!

Krukov:
I added some documentation, but I not sure that it is correct place. Also my english is not so well. Please review it. :confused:

Krukov:
@selwin Is everythikg ok here?

django_rq/queues.py
```diff
@@ -22,6 +23,28 @@ def get_commit_mode():
...
+    Return queue class from config or from RQ settings, otherwise return DjangoRQ
+    """
+    RQ = getattr(settings, 'RQ', {})
+    queue_class = DjangoRQ
+    if 'QUEUE_CLASS' in config:
+        queue_class = config.get('QUEUE_CLASS')
+    elif 'QUEUE_CLASS' in RQ:
+        queue_class = RQ.get('QUEUE_CLASS')
+
+    try:
```
selwin: Sorry for the late reply. Django >= 1.5 bundles `six` so we can remove the `try/except` statement here since the we support Django >= 1.5.

Krukov: Ok, I took it from the  `decorators.py`. So I will change it in a both files. Ok?

selwin:
@Krukov thanks for the PR!

Krukov:
@selwin  Thanks you. Can I ask you to release a new version?

PR #174, opened 2016-06-29
Added rqcronjobs management command to easily manage cron jobs from settings file

lechup:
Idea here:
https://github.com/ui/django-rq/issues/159#issuecomment-216763329

* created `rqcronjobs` management command
* added some tests
* added some documentation

To verify:
* does it make sense to use lazy for `RQ_CRONJOBS`
* are all variable names/function names sane?
* should we remove onlyt cronjobs while iterating over scheduler or it's ok to just remove eveything? how to check that?


selwin:
Thanks for the PR, @lechup . I think it's better if we implement a better way (one that doesn't require iterating through all jobs) to check whether a certain job is already scheduled.

This should be implemented in `rq-scheduler` so everyone can benefit from that (especially now that `rq-scheduler` supports cron jobs.

selwin:
I'm still thinking of how best we should implement this, so suggestions are welcome.

lechup:
Hm... Yeah. I only use cronjob in my projects (no enqueue_in/at) so iterating and by default remove old and add new jobs was the KISS'est sollution for me. For sure I should add a check there to filter only cron jobs... Or create a `get_cronjobs()` on scheduler API and use that one here in management command?

I guess by saying `It should be implemented in rq-scheduler` you mean some kind of generic API that would enable cronjobs management by setting file? That one I proposed is django one so I thought `django-rq` would be a better place...

I'm also open for discussion and proposals :)

PR #178, opened 2016-08-01
add number workers argument for better use of heroku dynos and autoreload for development

joesolly:


selwin:
@joesolly I think using Foreman to run multiple processes of RQ is a better approach than building this into DjangoRQ, take a look at the last section of this documentation: http://python-rq.org/patterns/.

This doesn't mean RQ shouldn't handle concurrency, of course. But if it does, it should be built into RQ and not DjangoRQ

joesolly:
@selwin I was recently working on a project where we scaled Heroku dynos in order to run multiple processes of RQ. Each dyno was being drastically underused and would have been more effectively used by having multiple processes/workers run in one dyno. While I agree that using Foreman to run multiple processes locally is probably ideal, it does not help us in Heroku.

As for RQ handling concurrency, looking at the documentation you linked to, RQ itself doesn't provide a built-in way to run workers. Their example shows how to create `run-worker.py`, which seems to be the starting off point for `management/commands/rqworker.py`. I would argue that process level management belongs where the workers are being created, not inside the workers themselves. (Which technically does manage processes but only one subprocess to actually perform the work before it dies to create a new process for a new job. In order to modify RQ to handle multiple workers, the main worker would need to be modified to continuously check the status of multiple children processes or it would need to have multiple workers that each manage their own process, neither of which, I think, are great solutions.)

I feel that having django-rq manage multiple workers is the appropriate location. I am open to further discussion if you would like.

P.S. Sorry to add multiple features in the same pull request, but as they were intertwined, I couldn't split them out into two features.

joesolly:
For anyone interested, these features have been added as a separately installable management command in https://github.com/istrategylabs/django-rq-wrapper.

PR #182, opened 2016-08-11
show job result on job django admin detail page.

mojeto:
Hi,
I created this pull request, because I want to be able to see job return value on job django admin detail page.

It works for job in any state. (It display None if no result yet)


selwin:
Thanks!

PR #183, opened 2016-08-11
adds "compatibility mode" for positional argument support across vers…

sbussetti:
…ions

See: https://docs.djangoproject.com/en/1.8/howto/custom-management-commands/

> 
> Changed in Django 1.8:
> Before Django 1.8, management commands were based on the optparse module, and positional arguments were passed in *args while optional arguments were passed in **options. Now that management commands use argparse for argument parsing, all arguments are passed in **options by default, unless you name your positional arguments to args (compatibility mode). You are encouraged to exclusively use **options for new commands.

### DIFF
--- a/django_rq/management/commands/rqenqueue.py
+++ b/django_rq/management/commands/rqenqueue.py
@@ -1,4 +1,7 @@
+from distutils.version import LooseVersion
+
 from django.core.management.base import BaseCommand
+from django.utils.version import get_version
 
 from django_rq import get_queue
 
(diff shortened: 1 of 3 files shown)


selwin:
Thanks for the PR!

Tests for this PR still fails. However, I'm thinking of dropping support for Django < 1.8. Do we still need this PR in that scenario?

sbussetti:
If you drop < 1.8 support we need a slightly different approach. I'll make
anotger PR and ensure tests are fixed.

-- 
Steve Bussetti

selwin:
Thanks!

selwin:
@sbussetti just a reminder to update this PR when you have the chance :)

sbussetti:
Fixed and tested.  I left the < 1.8 checks in for now tho.

selwin:
Thanks!

PR #184, opened 2016-08-22
Added exit code 1 in case of connection error

jeromer:
The use case is the following : 
- you run `rqworker` supervised by [supervisord](https://supervisord.org)
- you run any system operation on redis, like updating its configuration
- you restart redis
- the worker detects Redis is down and quits with the message 

> Registering death

but also with exit code 0 which is not in Supervisor"s unexpected exit codes.

Therefore supervisord will simply not restart the `rqworker` under supervision as it thinks it has stopped normally.

Exiting with exit code 1 in case of connection error fixes this.

### DIFF
--- a/django_rq/management/commands/rqworker.py
+++ b/django_rq/management/commands/rqworker.py
@@ -1,6 +1,7 @@
 import os
 import importlib
 import logging
+import sys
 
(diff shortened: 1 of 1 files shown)


selwin:
Thanks for the PR!

PR #188, opened 2016-09-06
Fixed --burst boolean arg of rqworker command

claudep:
While converting commands to argparse, the `--burst` flag to the `rqworker` command was badly transformed to a flag needed a parameter.


claudep:
Anyone for a review, this is an annoying bug right now.

selwin:
Sorry for the delay and thanks for the PR!

PR #192, opened 2016-09-29
rqenqueue fix and rqscheduler enhancement

vindemasi:
hello there are two little changes

1. fixed the rqenqueue command on --timeout option
2. added a --pid option to the rqscheduler command (adapted from the one in rqworker)

best regards


selwin:
Thanks for the PR!

PR #181, opened 2016-08-10
Queues are not passed via command line

hendi:
This fixes #180 on Django 1.10

PR #199, opened 2016-11-04
fix argument parsing to work on Django 1.8, 1.9, and 1.10

hendi:
This fixes a regression introduced by 2eb4afad which broke passing of queues on Django 1.9 (as reported in #181).

### DIFF
--- a/django_rq/tests/tests.py
+++ b/django_rq/tests/tests.py
@@ -192,7 +192,20 @@ class QueuesTest(TestCase):
         """
         Checks that passing queues via commandline arguments works
         """
-        self.assertRaises(KeyError, lambda: call_command('rqworker', 'some_queue'))
+        queue_names = ['django_rq_test', 'django_rq_test']
+        jobs = []
+        for queue_name in queue_names:
+            queue = get_queue(queue_name)
+            jobs.append({
+                'job': queue.enqueue(divide, 42, 1),
+                'finished_job_registry': FinishedJobRegistry(queue.name, queue.connection),
+            })
+
+        call_command('rqworker', *queue_names, burst=True)
+
+        for job in jobs:
+            self.assertTrue(job['job'].is_finished)
+            self.assertIn(job['job'].id, job['finished_job_registry'].get_job_ids())
 
     def test_get_unique_connection_configs(self):
         connection_params_1 = {
--- a/django_rq/tests/tests.py
+++ b/django_rq/tests/tests.py
@@ -188,6 +188,12 @@ class QueuesTest(TestCase):
         """
         self.assertRaises(ValueError, get_queues, 'default', 'test')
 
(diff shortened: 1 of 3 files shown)


pacahon:
@hendi Can you accept this ASAP and release on pypi? I have a problem without this fix.

```
$ python manage.py help rqworker

...
positional arguments:
  args
  queues                The queues to work on, separated by space
  args

optional arguments:
```

My workaround is to avoid `args` before `queues` (delete class attr `args`)

hendi:
Sorry @pacahon but I'm not the one with access rights to make releases on pypi.

@selwin Can you please create a new release with this bugfix? Otherwise django-rq is broken for Django < 1.10

selwin:
@hendi is it possible to add a management command test case for this? I'm worried about this breaking again.

hendi:
@selwin Good idea, I've added a new test.

pacahon:
@hendi In my case command didn't raise any exceptions.

What about parsing `rq.worker` debug output? 
Like I did here https://github.com/ui/django-rq/pull/201

selwin:
You could also run the management command in burst mode and see if calling the command runs the worker for the correct queues.

pacahon:
> You could also run the management command in burst mode and see if calling the command runs the worker for the correct queues.

That what I've did in #201, btw.

hendi:
@pacahon Which Django version are you using? The KeyError should get raised because `some_queue` isn't a valid queue.

I'll look into using the debug output for test, thanks for the feedback.

hendi:
I propose we do the following:

1. merge this
2. rebase #201 
3. adapt this test to the method used by #201

pacahon:
@hendi You are right, sorry. I meant the case, when I had valid queue name.
I agree, of course. Waiting for the others.

selwin:
@hendi yes, if you could please adapt this test to the method used in #201, witha small caveat, rather than checking for debug output, I'd rather us check that dummy jobs in the specified queues are processed. It seems cleaner that way.

pacahon:
@selwin Did I understand you correctly? Check this test, please.
 https://github.com/pacahon/django-rq/blob/3cd76817110c7d2e96a70097ad03cb027f74ea89/django_rq/tests/tests.py#L364

django_rq/tests/tests.py
```diff
@@ -192,7 +192,20 @@ def test_pass_queue_via_commandline_args(self):
         """
         Checks that passing queues via commandline arguments works
         """
-        self.assertRaises(KeyError, lambda: call_command('rqworker', 'some_queue'))
+        queue_names = ['django_rq_test', 'django_rq_test']
```
pacahon: typo?

selwin:
Thanks @hendi !

PR #201, opened 2016-11-08
WIP tests for `rqworker` command positional arguments

pacahon:
It should fail right now (at least on Django 1.9.x) due to wrong argument parsing. Needs review.


selwin:
Thanks for the PR!. This is fixed in https://github.com/ui/django-rq/pull/199

pacahon:
I believe in fail-fast principle, so it will be good to add tests from this PR too, but it's all up to you, of course.

PR #213, opened 2017-01-27
consider one and only one failed queue for each distinct redis connection

### LINKED ISSUE #175: Mulitple failed Queues (if different time-outs)
If I define more than a queue each with different DEFAULT_TIMEOUT than I obtain more than one "failed" queue in the stats view

I think the reason is here:
https://github.com/ui/django-rq/blob/master/django_rq/queues.py#L201

The connection_configs is different because of "DEFAULT_TIMEOUT" item (which is not strictly a connection parameter)

At the moment I used this workaround: specify exactly the same items for each Queue, ... but it is not exactly what I would want

See also #97

depaolim:
[fix #175]

### DIFF
--- a/django_rq/queues.py
+++ b/django_rq/queues.py
@@ -218,6 +218,7 @@ def get_unique_connection_configs(config=None):
 
     connection_configs = []
     for key, value in config.items():
+        value = filter_connection_params(value)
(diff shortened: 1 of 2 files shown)


selwin:
Thanks!

PR #215, opened 2017-01-30
Each queue should be able to have its custom_exception or None.

fchevitarese:
Each queue should be able to have its custom_exception or None. 

Make a global custom_exception may lead us to a non desired behavior on other queues.

### DIFF
--- a/django_rq/management/commands/rqworker.py
+++ b/django_rq/management/commands/rqworker.py
@@ -77,7 +77,8 @@ class Command(BaseCommand):
                 queues,
                 connection=queues[0].connection,
                 name=options['name'],
-                exception_handlers=get_exception_handlers() or None,
+                exception_handlers=get_exception_handlers(
+                    queues[0].exception_handlers) or None,
                 default_worker_ttl=options['worker_ttl']
             )
 
--- a/django_rq/queues.py
+++ b/django_rq/queues.py
@@ -49,6 +49,7 @@ class DjangoRQ(Queue):
     def __init__(self, *args, **kwargs):
         autocommit = kwargs.pop('autocommit', None)
         self._autocommit = get_commit_mode() if autocommit is None else autocommit
+        self.exception_handlers = kwargs.pop('exception_handlers', None)
 
         super(DjangoRQ, self).__init__(*args, **kwargs)
 
@@ -138,9 +139,13 @@ def get_queue(name='default', default_timeout=None, async=None,
         default_timeout = QUEUES[name].get('DEFAULT_TIMEOUT')
     if queue_class is None:
         queue_class = get_queue_class(QUEUES[name])
+
+    exception_handlers = QUEUES[name].get('EXCEPTION_HANDLERS', None)
+
     return queue_class(name, default_timeout=default_timeout,
                        connection=get_connection(name), async=async,
-                       autocommit=autocommit)
+                       autocommit=autocommit,
+                       exception_handlers=exception_handlers)
 
 
 def get_queue_by_index(index):
--- a/django_rq/workers.py
+++ b/django_rq/workers.py
@@ -2,17 +2,27 @@ from rq import Worker
 from rq.utils import import_attribute
 
 from .queues import get_queues
-from .settings import EXCEPTION_HANDLERS
 
 
-def get_exception_handlers():
-    """
-    Custom exception handlers could be defined in settings.py:
-    RQ = {
-        'EXCEPTION_HANDLERS': ['path.to.handler'],
+def get_exception_handlers(exception_handlers):
+    """Custom exception handler defined in QUEUE settings:
+    RQ_QUEUES = {
+        'default': {
+            'HOST': 'localhost',
+            'PORT': 6379,
+            'DB': 0,
+            'PASSWORD': '',
+            'DEFAULT_TIMEOUT': 360,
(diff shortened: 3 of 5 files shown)


fchevitarese:
How can i run tests in my machine?

selwin:
I agree that this is a good idea. However, I'd like to keep it backwards compatible if we can. Could you please change your PR to make `exception_handlers` fallback to `RQ_EXCEPTION_HANDLERS`?

fchevitarese:
Will do ;)

fchevitarese:
i'm sorry, i was really busy those days ;) 

Take a look and see if it fits ^^

django_rq/workers.py
```diff
@@ -2,24 +2,42 @@
...
+            'DEFAULT_TIMEOUT': 360,
+            'EXCEPTION_HANDLERS': [
+                'test_exception_handler.my_custom_exception'
+            ],
+        }
     }
     """
-    return [import_attribute(path) for path in EXCEPTION_HANDLERS]
+    try:
+        from django.settings import RQ_EXCEPTION_HANDLERS
```
selwin: `RQ_EXCEPTION_HANDLERS` should only be a fallback. So we only try to import it if `exception_handlers` is `None`.

fchevitarese: Ok, i will rewrite asap ;)

fchevitarese:
Done, could you take a look if it's ok? 

Cheers!

django_rq/workers.py
```diff
@@ -1,25 +1,40 @@
...
 from .settings import EXCEPTION_HANDLERS
+from .queues import get_queues
 
 
-def get_exception_handlers():
-    """
-    Custom exception handlers could be defined in settings.py:
-    RQ = {
-        'EXCEPTION_HANDLERS': ['path.to.handler'],
+def get_exception_handlers(exception_handlers):
```
selwin: I think it's better to change the function signature to `get_exception_handlers(queue_name)`

django_rq/queues.py
```diff
@@ -49,6 +49,7 @@ class DjangoRQ(Queue):
     def __init__(self, *args, **kwargs):
         autocommit = kwargs.pop('autocommit', None)
         self._autocommit = get_commit_mode() if autocommit is None else autocommit
+        self.exception_handlers = kwargs.pop('exception_handlers', None)
```
selwin: There's no need to store `exception_handlers` as property in `DjangoRQ`

selwin:
@fchevitarese could you also fix the failing tests?

django_rq/queues.py
fchevitarese: Ok. About the failing tests, idk how to fix it because, once the exception handler comes in, i couldn't find how to put them in failed again if they fail. 

That's the problem with the test. The failed job never go to failed after a exception_handlers takes in. :( 

I would like some help about it , if possible. 

Thanks! ^^

fchevitarese:
@selwin , the only way i found to fix the testing was to check if exception_handler exists, and if not, start a Worker without the parameter. 
Even when there's no [exception_handler] it was raising the error when trying to get the failing job to requeue. 

I'm not sure why, and i couldn't find a way to fix this problem in other way. 

About the property "self.exception_handlers = kwargs.pop('exception_handlers', None)" i removed, but started to face a lot of errors, so i put it back. 

Hope it's all good!

django_rq/management/commands/rqworker.py
```diff
@@ -77,7 +77,8 @@ def handle(self, *args, **options):
                 queues,
                 connection=queues[0].connection,
                 name=options['name'],
-                exception_handlers=get_exception_handlers() or None,
+                exception_handlers=get_exception_handlers(
```
selwin: I don't think this is right. The function is defined as `get_exception_handlers(queue_name)` but you're passing in `queue.exception_handlers` here.

django_rq/queues.py
selwin: You can simply write a test for `get_queue('default')` and ensure that `queue.exception_handlers` contains the exception handler you specified in settings.

django_rq/queues.py
```diff
@@ -138,9 +139,13 @@ def get_queue(name='default', default_timeout=None, async=None,
...
     if queue_class is None:
         queue_class = get_queue_class(QUEUES[name])
+
+    exception_handlers = QUEUES[name].get('EXCEPTION_HANDLERS', None)
+
     return queue_class(name, default_timeout=default_timeout,
                        connection=get_connection(name), async=async,
-                       autocommit=autocommit)
+                       autocommit=autocommit,
+                       exception_handlers=exception_handlers)
```
selwin: You should also import `exception_handlers` before passing it into `__init__`

django_rq/queues.py
selwin: Sorry, the test should be for `get_worker()` because only workers have `exception_handlers`. `Queue` objects have no `exception_handlers` in RQ.

django_rq/queues.py
selwin: @fchevitarese reminder to update this PR when you're free :)

fchevitarese:
Hi! I'm sorry I've been very busy. Will do when I can

Em 22/07/2017 21:54, "Selwin Ong" <notifications@github.com> escreveu:

Issue #216, opened 2017-02-07
If the worker loses the DB connection it is not able to reconnect and you must restart it


depaolim:
If the worker loses the DB connection it isn't able to reconnect and you must restart it

Workers should adopt the same DB connection/reconnection policy that Django uses in his request/response cycle

Every time a django-rq worker performs a job, the worker should assure the database-orm-connections are in a valid state.
This policy is enforced by django in its request/response cycle via the function ```close_old_connections```

Look at:
https://github.com/django/django/blob/master/django/db/__init__.py

Similary, in django-rq worker, the ```close_old_connections``` should be called before and after the execution of each job

(see also #49 for some preliminary thoughts)

depaolim:
@selwin please, feel free to tell me your opinion

I set ```DjangoWorker``` as the default django-rq worker.
The rationale behind that decision is that any django-rq worker should be "aware" of the Django ORM connections and manage them according to Django ORM default pattern: reconnect if not valid.
(In a similar way ```DjangoRQ``` is a rq.Queue which is "aware" of Django transactions)

@aaugustin sorry to bother you
you are the author of django persistent-connections (see https://github.com/django/django/commit/2ee21d9f0d9eaed0494f3b9cd4b5bc9beffffae5).
You are perfectly placed to tell me if the approach is correct from the point of view of Django

TIA to both

aaugustin:
Yes I think it would make sense for the RQ worker to handle database connections around tasks like Django does around HTTP requests with:

It boils down to executing:

```
from django.db import connections
for conn in connections.all():
    conn.close_if_unusable_or_obsolete()
```

before and after each task.

See `django/db/__init__.py` for details.

depaolim:
Ok, thank you!

This is the snippet I used in my PR (https://github.com/ui/django-rq/pull/217/files#diff-894c6dcfb79147f4bb0823044b78db19):

```
from django.db import close_old_connections, reset_queries

class DjangoWorker(rq.Worker):
    def perform_job(self, job, queue):
        reset_queries()
        close_old_connections()
        try:
            super(DjangoWorker, self).perform_job(job, queue)
        finally:
            close_old_connections()
```

(I added also a call to ```reset_queries``` as seen in ```django/db/__init__.py```)

depaolim:
The problem is another.
My previous solution is not sufficient.
If ```CONN_MAX_AGE``` is not zero in settings DATABASES, than the worker is unable to reconnect

The picture, as far as I can understand, is the following:

[...]

PR #217, opened 2017-02-07
the new DjangoWorker manages django.db.connections cleanup with the same policy of Django [see #216]

depaolim:
see #216

### DIFF
--- a/django_rq/workers.py
+++ b/django_rq/workers.py
@@ -1,8 +1,20 @@
-from rq import Worker
+import rq
 from rq.utils import import_attribute
 
 from .queues import get_queues
 from .settings import EXCEPTION_HANDLERS
+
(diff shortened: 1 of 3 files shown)


depaolim:
Please, don't merge this PR. See #216

django_rq/workers.py
```diff
@@ -1,9 +1,21 @@
...
+
+
+class DjangoWorker(rq.Worker):
+    def perform_job(self, job, queue):
+        reset_queries()
+        close_old_connections()
+        try:
+            super(DjangoWorker, self).perform_job(job, queue)
+        finally:
+            close_old_connections()
```
selwin: Do we really need to call `close_old_connection()` after `perform_job`?

PR #219, opened 2017-02-10
new command rqclean [fix #218]

### LINKED ISSUE #218: command to cleanup a queue
Create a new command ```rqclean``` that removes all queue jobs for the queue specified in the command line

depaolim:


django_rq/management/commands/rqclean.py
```diff
@@ -0,0 +1,25 @@
+from django.core.management.base import BaseCommand
+
+from django_rq import get_queue
+
+
+class Command(BaseCommand):
```
selwin: Given the nature of this command, I think we should prompt the user whether really wants to remove all jobs from the named queue. 

Would `rqempty` or `rqflush` be more descriptive than `rqclean`?

selwin:
Oh, and could you please document this management command?

depaolim:
Actions:
* renamed ``rqclean`` to ``rqflush``
* asks for confirmation
* documented in README

... unfortunatelly I'have made some mess with the commits.
The final differences should be ok (file changes are the minimal ones). But there are some spurious commit.
So, please, let me know if you prefer I throw away this PR completelly and create a new one

depaolim:
no need to implement anything, use:
```rq empty <queue-name>```

PR #220, opened 2017-02-15
rqworker cleans up any DB connection before running any job [fix #216]

### LINKED ISSUE #216: If the worker loses the DB connection it is not able to reconnect and you must restart it

depaolim:


selwin:
Thanks for the PR and the extensive test cases!

PR #224, opened 2017-03-09
Fix view paging for registry-based job lists

### LINKED ISSUE #223: Paging through Registry jobs does not work correctly
In the Django RQ views, when there are multiple pages of jobs in any of the RQ registries (finished jobs, started jobs, deferred jobs), paging through these jobs in the UI does not work correctly.

This is easily reproducible in any system where there is moderately high job throughput (say a few hundred jobs in the `FinishedJobRegistry` at any given point in time) by examining the finished jobs view for a queue. The first page displays fine, but subsequent pages display only one job (for page 2) or none at all (for all subsequent pages), though the number of pages is listed correctly. Here are pages 2 and 3 for the finished jobs in my `default` queue on one of our production systems right now:

image

image

This is happening because the function signatures in RQ for `Queue.get_job_ids` / `Queue.get_jobs` and `BaseRegistry.get_job_ids` are different. Below are the relevant signatures from RQ.

https://github.com/nvie/rq/blob/master/rq/queue.py#L121: 
```python
def get_job_ids(self, offset=0, length=-1):
```

https://github.com/nvie/rq/blob/master/rq/queue.py#L131:
```python
def get_jobs(self, offset=0, length=-1):
```

https://github.com/nvie/rq/blob/master/rq/registry.py#L57
```
def get_job_ids(self, start=0, end=-1):
```

So `Queue.get_jobs` and `Queue.get_job_ids` expect an offset and a length when requesting jobs, and `BaseRegistry.get_job_ids` expects a start point (same thing as offset) and an end point (not at all the same as a length), but Django RQ uses a length when calling both - e.g. here's the line which gets the jobs for a particular page on the `FinishedJobsRegistry` for a queue:

https://github.com/ui/django-rq/blob/master/django_rq/views.py#L107
```python
job_ids = registry.get_job_ids(offset, items_per_page)
```

The obvious way to fix this is to change the calls in Django RQ to correctly match the signature for RQ's `BaseRegistry.get_job_ids`. There is an argument to be made that RQ's `BaseRegistry.get_job_ids` signature should be changed to be consistent with `Queue.get_job_ids`, but that could be a breaking change for other downstream apps or libraries that are depending on the current signature. @selwin, if you agree with changing this in Django RQ, I will make a PR.

smaccona:
Closes #223.

### DIFF
--- a/django_rq/views.py
+++ b/django_rq/views.py
@@ -104,7 +104,7 @@ def finished_jobs(request, queue_index):
         last_page = int(ceil(num_jobs / items_per_page))
         page_range = range(1, last_page + 1)
         offset = items_per_page * (page - 1)
-        job_ids = registry.get_job_ids(offset, items_per_page)
+        job_ids = registry.get_job_ids(offset, offset + items_per_page - 1)
 
         for job_id in job_ids:
             try:
@@ -143,7 +143,7 @@ def started_jobs(request, queue_index):
         last_page = int(ceil(num_jobs / items_per_page))
         page_range = range(1, last_page + 1)
         offset = items_per_page * (page - 1)
-        job_ids = registry.get_job_ids(offset, items_per_page)
+        job_ids = registry.get_job_ids(offset, offset + items_per_page - 1)
 
         for job_id in job_ids:
             try:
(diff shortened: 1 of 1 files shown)


selwin:
Thanks!

PR #227, opened 2017-03-20
Use a standalone file for the changelog

jayfk:
First of all, thanks for writing a changelog 👍 

Would it be possible to use a standalone file for that? This would make it easier for automated tools to pick it up.


selwin:
Thanks!

PR #233, opened 2017-05-04
Fix #231 Improves interface response

### LINKED ISSUE #231: Django-rq slow interface
I have 4 worker servers orchestrated by docker swarm, running more than 600 workers. I noticed the interface `rq_home` got slow because of the number of workers.

When running the function `django_rq.views.stats` alone, it took me 10 seconds to respond. This part in special seems slower

    all_workers = Worker.all(connection=connection)

This is really annoying me because we depend on the interface. Is there any way to get through this slowness?

lucastamoios:
In the view `stats`, the query `Worker.all()` is applied to every queue
in the settings, even those which have the same configuration and, thus,
would give the same response.

This commit aims to avoid all unnecessary queries by calling a
preprocessor.

### DIFF
--- a/django_rq/views.py
+++ b/django_rq/views.py
@@ -14,15 +14,15 @@ from rq.job import Job
 from rq.registry import (DeferredJobRegistry, FinishedJobRegistry,
                          StartedJobRegistry)
 
-from .queues import get_connection, get_queue_by_index
+from .queues import get_connection, get_queue_by_index, filter_connection_params 
 from .settings import QUEUES_LIST
-from .workers import collect_workers_per_configuration
+from .workers import collect_workers_by_connection, get_all_workers_by_configuration
 
 
 @staff_member_required
 def stats(request):
     queues = []
-    workers_collections = collect_workers_per_configuration(QUEUES_LIST)
+    workers_collections = collect_workers_by_connection(QUEUES_LIST)
     for index, config in enumerate(QUEUES_LIST):
 
         queue = get_queue_by_index(index)
@@ -43,7 +43,10 @@ def stats(request):
 
         else:
             connection = get_connection(queue.name)
-            all_workers = _get_all_workers(config['connection_config'], workers_collections)
+            all_workers = get_all_workers_by_configuration(
+                config['connection_config'],
+                workers_collections
+            )
             queue_workers = [worker for worker in all_workers if queue in worker.queues]
             queue_data['workers'] = len(queue_workers)
 
@@ -58,12 +61,6 @@ def stats(request):
 
     context_data = {'queues': queues}
     return render(request, 'django_rq/stats.html', context_data)
-
-
-def _get_all_workers(config, workers_collections):
-    for collection in workers_collections:
-        if config == collection['config']:
-            return collection['all_workers']
 
 
 @staff_member_required
--- a/django_rq/views.py
+++ b/django_rq/views.py
@@ -16,12 +16,13 @@ from rq.registry import (DeferredJobRegistry, FinishedJobRegistry,
 
 from .queues import get_connection, get_queue_by_index
 from .settings import QUEUES_LIST
+from .workers import collect_workers_per_configuration
 
 
 @staff_member_required
 def stats(request):
     queues = []
-    workers_collections = _collect_workers_per_configuration(QUEUES_LIST)
+    workers_collections = collect_workers_per_configuration(QUEUES_LIST)
     for index, config in enumerate(QUEUES_LIST):
 
         queue = get_queue_by_index(index)
@@ -57,22 +58,6 @@ def stats(request):
 
     context_data = {'queues': queues}
     return render(request, 'django_rq/stats.html', context_data)
-
-
-def _collect_workers_per_configuration(queue_list):
-    """Collects, into a list, dictionaries of connections_config and its
-    workers
-    """
-    workers_collections = []
-    for item in queue_list:
-        if item['connection_config'] not in [c['config'] for c in workers_collections]:
-            connection = get_connection(item['name'])
-            collection = {
-                'config': item['connection_config'],
-                'all_workers': Worker.all(connection=connection)
-            }
-            workers_collections.append(collection)
-    return workers_collections
 
 
 def _get_all_workers(config, workers_collections):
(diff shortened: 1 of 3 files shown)


django_rq/views.py
```diff
@@ -58,6 +59,28 @@ def stats(request):
     return render(request, 'django_rq/stats.html', context_data)
 
 
+def _collect_workers_per_configuration(queue_list):
```
selwin: This is an excellent optimization and should really help if you have many queues in the same Redis DB.

django_rq/views.py
```diff
@@ -58,6 +59,28 @@ def stats(request):
...
+    """Collects, into a list, dictionaries of connections_config and its
+    workers
+    """
+    workers_collections = []
```
selwin: Please change this to a dict instead, keyed by `connection_config` with each key having a list of workers as its value.

django_rq/views.py
selwin: To keep things tidy, could you please move this function to `workers.py`?

django_rq/views.py
```diff
@@ -58,6 +59,28 @@ def stats(request):
...
+    for item in queue_list:
+        if item['connection_config'] not in [c['config'] for c in workers_collections]:
```
selwin: We should call `filter_connection_params` on each `connection_config` to make this even more efficient. Queues may have different configs although they point to the same Redis connection. For example:
```python
RQ_QUEUES = {
    'default': {
        'HOST': 'localhost',
        'PORT': 6379,
        'DB': 0,
        'PASSWORD': 'some-password',
        'DEFAULT_TIMEOUT': 360,
    },
     'low': {
        'HOST': 'localhost',
        'PORT': 6379,
        'DB': 0,
        'PASSWORD': 'some-password',
        'DEFAULT_TIMEOUT': 60, # This timeout causes the connection_configs to be different even though they use the same Redis DB
    },
}```

File: django_rq/views.py (hunk 2 above)
Comment (lucastamoios) [inline django_rq/views.py]: Correct me see if I understood it wrong, but `connection_config` is a `dict`, which in an unhashable type, and I cannot use it as a key.

File: django_rq/views.py (hunk 3 above)
Comment (lucastamoios) [inline django_rq/views.py]: Done. ce234020f46cc656f7d8ec5cff476672f74a744d

File: django_rq/views.py (hunk 1 above)
Comment (lucastamoios) [inline django_rq/views.py]: Done. ce234020f46cc656f7d8ec5cff476672f74a744d

File: django_rq/workers.py [hunk 4]
```diff
@@ -23,3 +23,20 @@ def get_worker(*queue_names):
     return Worker(queues,
                   connection=queues[0].connection,
                   exception_handlers=get_exception_handlers() or None)
+
+
+def collect_workers_per_configuration(queue_list):
```
selwin: Can we rename this to `collect_workers_by_connection(queues)`? I think it's more elegant if this returns a dictionary of workers keyed by its connection config like this:
```python
{
    CONFIG_1: [worker1],
    CONFIG_2: [worker2],
}
```
We can then use this in `views.py` as such:
```python
workers_dict = collect_workers_by_connection(QUEUES_LIST)

for index, config in enumerate(QUEUES_LIST):
    all_workers = workers_dict[config['connection_config']]
```

lucastamoios: If I understood it well, `CONFIG_1` and `CONFIG_2` are dictionaries, and when I do 
```python
{
    CONFIG_1: [worker1],
    CONFIG_2: [worker2],
}
```
What I am really doing is using a dict as a hash to another dict. Well, it gives us a `TypeError`, because dict is an *unhashable type*.

Let me know if did not understand it well.

In relation to the names, I agree with you, but I will wait to submit when this thread close.

selwin: I forgot that dict is unhashable. To make it hashable, you can turn the dict it into a string e.g: `dictionary[str(connection_params)] = []`

django_rq/views.py
```diff
@@ -58,6 +60,12 @@ def stats(request):
     return render(request, 'django_rq/stats.html', context_data)
 
 
+def _get_all_workers(config, workers_collections):
+    for collection in workers_collections:
```
selwin: Don't forget to call `filter_connection_params` here, or it won't work if you have connections with `DEFAULT_TIMEOUT`.

django_rq/workers.py
selwin: Could you please add a test case for `collect_workers_by_connection` and `_get_all_workers`?

django_rq/views.py
lucastamoios: I call it on `workers.py`, inside `collect_workers_by_connection`. I thought would be better.
```python
workers_collections = []
for item in queue_list:
    connection_params = filter_connection_params(item['connection_config'])
# ...
```

django_rq/workers.py
lucastamoios: Seems a bit kludge to me. As the dict built-in type does not guarantee any particular order of the keys, we cannot rely on this property to assert any connection is or not stored.

django_rq/views.py
lucastamoios: Thanks, I did not figure out I would need it here before...

django_rq/workers.py
selwin: Ah, right. We can turn the keys into frozensets: https://stackoverflow.com/questions/13264511/typeerror-unhashable-type-dict but as you said it does look kludgy.

Let's keep it the way it is now then.

django_rq/views.py
selwin: @lucastamoios can you update the line here?

django_rq/workers.py
selwin: One more thing, do you mind documenting this part more extensively, including an example of how workers_collections look like?

django_rq/views.py
lucastamoios: I took the liberty to put the function `_get_all_workers` in `django_rq/workers.py` as `get_all_workers_by_configuration()`. I think it got more expressive.

django_rq/workers.py
lucastamoios: Just remembering that the changes were implemented and are waiting for approval. :)

lucastamoios:
@selwin Is there anything else I need to do in order to approve this pull request?

selwin:
Sorry, haven't had the time to look at this yet. Will let you know soon.

selwin:
Thanks!

PR #235, opened 2017-06-07
patterns function removed from django

chaffeqa:
deprecated since 1.8, removed in 1.10


chaffeqa:
Oops meant this to be on our fork

PR #237, opened 2017-07-21
Broken Dependency Fix

chaffeqa:
When a job depends on another that is deleted, it becomes stranded, cannot be requeued, and ends up serving a 500.

This fix allows for:

* Rescuing
* Fixing job to have no dependency
* Moving it to failed queue where it can be requeued

I added to [the rq handling discussion](https://github.com/nvie/rq/pull/427#issuecomment-317022979)

### DIFF
--- a/django_rq/templates/django_rq/job_detail.html
+++ b/django_rq/templates/django_rq/job_detail.html
@@ -97,7 +97,7 @@
                 <label class="required">Dependency:</label>
                 <div class="data">
                   <a href = "{% url 'rq_job_detail' queue_index job.dependency.id %}">
-                    {{ job.dependency.func_name }}
+                    {{ job.dependency.id }} {{ job.dependency.func_name }}
                   </a>
                 </div>
             </div>
--- a/django_rq/templates/django_rq/job_detail.html
+++ b/django_rq/templates/django_rq/job_detail.html
@@ -68,12 +68,14 @@
             </div>
         </div>
 
+        {% if job.enqueued_at %}
         <div class="form-row">
             <div>
                 <label class="required">Queued:</label>
                 <div class="data">{{ job.enqueued_at|to_localtime }}</div>
             </div>
         </div>
+        {% endif %}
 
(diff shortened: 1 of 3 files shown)


django_rq/views.py
```diff
@@ -218,6 +219,29 @@ def job_detail(request, queue_index, job_id):
         job = Job.fetch(job_id, connection=queue.connection)
     except NoSuchJobError:
         raise Http404("Couldn't find job with this ID: %s" % job_id)
+    # job should have dependency but it got deleted we need to cleanup
+    try:
+        job.dependency
+    except Exception as e:
```
selwin: Can we be more explicit as to what exception we want to catch?

django_rq/templates/django_rq/job_detail.html
```diff
@@ -89,6 +91,19 @@
             </div>
         </div>
 
+        {% if job.dependency %}
+        <div class="form-row">
+            <div>
+                <label class="required">Dependency:</label>
+                <div class="data">
+                  <a href = "{% url 'rq_job_detail' queue_index job.dependency.id %}">
+                    {{ job.dependency.func_name }}
```
selwin: I think we should also display the `job.id` here, something like this would be great `{{ job.dependency.id }} ({{ job.dependency.func_name }}) `

django_rq/views.py
```diff
@@ -218,6 +219,29 @@ def job_detail(request, queue_index, job_id):
...
+        # remove the dependency association
+        job.connection.srem(job.dependents_key_for(dependency_id), job.id)
```
selwin: The way this tries to use Redis commands to directly cleanup dependent jobs is making me nervous. I think we should first add `job.delete_dependents()` to RQ so it can be called here.

selwin:
@chaffeqa reminder to update this PR when you have the time :)

django_rq/views.py
kingbuzzman: :+1:

PR #241, opened 2017-08-07
Worker stats

selwin:


wjdp:
Any chance of a release for this? (Got an issue that worker stats would be rather useful for)

PR #245, opened 2017-09-23
Optimise requeue all - resolves #244

### LINKED ISSUE #244: Memory exhaustion when doing "Requeue all" with a large number of failed jobs
I think I hit a memory exhaustion issue when trying to requeue ~90,000 failed jobs

It looks like this could be avoided if django_rq.views.requeue_all used a generator to iterate through each job rather than loading all into memory at the start.

therefromhere:
Only load job ids when doing requeue all, instead of instantiating job objects, to save resources.

Also added gitignore of pycharm project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@
 dist/
 django_rq.egg-info/
 
+# pycharm
+.idea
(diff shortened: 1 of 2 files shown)


selwin:
Thanks!

PR #246, opened 2017-09-23
Bump test requirements

therefromhere:
Also 

- gitignore pycharm config
- Use Django's six (fixes issue that six was missing from test requirements)
- Pycharm optimise imports on `integration_test/_tests.py`


selwin:
Thanks!

PR #251, opened 2017-10-12
Fix issue displaying deferred queue

### LINKED ISSUE #250: Cannot view deferred queue
When trying to view the deferred queue, I'm getting an error is encountered in the to_localtime in django_rq/templatetags/django_rq.py:14, seemingly because the enqueued_at attribute for a deferred job is None.

Relevant error sections:
Internal Server Error: /django-rq/queues/0/deferred/

AttributeError at /django-rq/queues/0/deferred/
'NoneType' object has no attribute 'replace'

Relevant line of the traceback:
File "/srv/app/env/lib/python3.6/site-packages/django_rq/templatetags/django_rq.py" in to_localtime
  14.     utc_time = time.replace(tzinfo=timezone.utc)

django_rq version: django-rq==0.9.6
django version: Django==1.11.6

Steps to reproduce:
Enqueue a new job
Create another new job, which depends on the first job
Try to view the deferred queue

Hovercross:
A simple fix to allow object passed into to_localtime to be None and not cause an exception to be thrown. Specifically fixes an issue that I was seeing when viewing the deferred queue.

Closes #250

### DIFF
--- a/django_rq/templates/django_rq/job_detail.html
+++ b/django_rq/templates/django_rq/job_detail.html
@@ -61,20 +61,24 @@
             </div>
         </div>
 
+        {% if job.created_at %}
         <div class="form-row">
             <div>
                 <label class="required">Created:</label>
                 <div class="data">{{ job.created_at|to_localtime|date:"Y-m-d, H:i:s" }}</div>
             </div>
         </div>
+        {% endif %}
 
+        {% if job.enqueued_at %}
         <div class="form-row">
             <div>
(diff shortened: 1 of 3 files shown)


selwin:
Could you please update this PR to make this check at the template level? Simply render `-` if `job.enqueued_at` is `None`.

I don't think modifying `to_localtime` to handle `None` is correct - it expects a datetime to be passed in.

selwin:
Thanks!

PR #253, opened 2017-10-18
Add support for sentry parameter

efi-mk:
Hello,

Added same support rqworker original command has for sentry.
Use ``sentry-dsn`` parameter when running rqworker. ``./manage.py rqworker sentry-dsn=https://*****@sentry.io/222222``

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -233,7 +233,7 @@ modifies the default admin template.
 
 Configuring Sentry
 -------------------
-Use ``sentry-dsn`` parameter when running rqworker. ``./manage.py rqworker sentry-dsn=https://*****@sentry.io/222222`` 
+Use ``sentry-dsn`` parameter when running rqworker. ``./manage.py rqworker --sentry-dsn=https://*****@sentry.io/222222`` 
 
 Configuring Logging
 -------------------
--- a/README.rst
+++ b/README.rst
@@ -231,6 +231,9 @@ You can also add a link to this dashboard link in ``/admin`` by adding
 override the default admin template so it may interfere with other apps that
 modifies the default admin template.
 
+Configuring Sentry
+-------------------
+Use ``sentry-dsn`` parameter when running rqworker. ``./manage.py rqworker sentry-dsn=https://*****@sentry.io/222222`` 
 
 Configuring Logging
 -------------------
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@ from setuptools import setup
 
 setup(
     name='django-rq',
-    version='0.9.6',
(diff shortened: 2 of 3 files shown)


README.rst
```diff
@@ -231,6 +231,9 @@ You can also add a link to this dashboard link in ``/admin`` by adding
 override the default admin template so it may interfere with other apps that
 modifies the default admin template.
 
+Configuring Sentry
+-------------------
+Use ``sentry-dsn`` parameter when running rqworker. ``./manage.py rqworker sentry-dsn=https://*****@sentry.io/222222``
```
deejay1: shouldn't this be `./manage.py rqworker --sentry-dsn=https://*****@sentry.io/222222` ?

setup.py
```diff
@@ -14,7 +14,7 @@
     zip_safe=False,
     include_package_data=True,
     package_data={'': ['README.rst']},
-    install_requires=['django>=1.8.0', 'rq>=0.5.5'],
+    install_requires=['django>=1.8.0', 'rq>=0.5.5','raven>=6.1.0'],
```
deejay1: I don't think this should be a hard dependency, it isn't required by RQ itself...

README.rst
efi-mk: Correct. Will fix.

setup.py
efi-mk: Agree. Will fix

efi-mk: Do you think I should specify it as optional via extras_require. See http://peak.telecommunity.com/DevCenter/setuptools#declaring-extras-optional-features-with-their-own-dependencies ? I can instead locally import and in case of an exception write a blurb to the console explaining the situation. What do you think ?

deejay1: I'd go with extras_require, as for the message I think the import error should be descriptive enough, rq also doesn't do any special information in case it's missing.

efi-mk:
Ok, fixed and pushed

efi-mk:
Any update ?

efi-mk:
Great, when do plan to merge ? When do you plan to release a new version ?

deejay1:
@efi-mk Sorry, I'm not the maintainer here, although we use django-rq quite extensively here...

pnuckowski:
+1

efi-mk:
@deejay1  oh sorry. Who should handle it ?

efi-mk:
@selwin any idea who is responsible for merging back pull requests ?

selwin:
I’ll take a look at this sometime in the next few weeks, after RQ 0.9.0 is released.

selwin:
Thanks!

Issue #255, opened 2017-11-09
Basic monitoring statistics for queues without all the HTML


SeiryuZ:
Really useful for statistics to be sent to monitoring services like datadog

selwin:
Fixed in https://github.com/rq/django-rq/pull/256

PR #256, opened 2017-11-09
Basic statistics refactor

SeiryuZ:
Initial stab at #255 

In this:
* Refactor stats view functionalities into utils.get_statistics
* Create CLI for statistics
* Add another view to retrieve statistics as JSON

demo-django-cli-dashboard


django_rq/urls.py
```diff
@@ -5,6 +5,10 @@
 urlpatterns = [
     url(r'^$',
         views.stats, name='rq_home'),
+    url(r'^stats.json/$',
```
selwin: Can you please combine the two `stats.json` views into one with optional `token` parameter?

django_rq/settings.py
```diff
@@ -22,3 +22,6 @@
 
 # Get exception handlers
 EXCEPTION_HANDLERS = getattr(settings, 'RQ_EXCEPTION_HANDLERS', [])
+
+# Token for querying statistics
+STATS_TOKEN = getattr(settings, 'RQ_STATS_TOKEN', 'asdasd')
```
selwin: Should default to empty string.

django_rq/views.py
```diff
@@ -4,64 +4,36 @@
...
-        queues.append(queue_data)
-
-    context_data = {'queues': queues}
-    return render(request, 'django_rq/stats.html', context_data)
+    return render(request, 'django_rq/stats.html', get_statistics())
+
+
+def token_stats_json(request, token):
+    if token != STATS_TOKEN:
+        raise Http404
```
selwin: Return a descriptive error message: `{"description": "Please configure API_TOKEN in settings.py before accessing this view"}`

PR #259, opened 2017-11-21
Show New statistics from rq 0.9.x on worker details page

SeiryuZ:
rq 0.9.x introduce some new worker statistics. This PR shows that in the django admin worker details page.

Also, added `oldest_queued_timestamp` to queue statistics

### DIFF
--- a/django_rq/templates/django_rq/stats.html
+++ b/django_rq/templates/django_rq/stats.html
@@ -24,7 +24,7 @@
                 <tr>
                     <th>Name</th>
                     <th>Queued Jobs</th>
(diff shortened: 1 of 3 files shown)


django_rq/templates/django_rq/stats.html
```diff
@@ -24,6 +24,7 @@
                 <tr>
                     <th>Name</th>
                     <th>Queued Jobs</th>
+                    <th>Oldests Queued Timestamp</th>
```
selwin: Rename to `Oldest Queued Job`

PR #263, opened 2017-11-28
Auto-reloading worker

### LINKED ISSUE #262: Use django's autoreload feature
Autoreloading has been a long-requested RQ feature. Since it doesn't seem it's going to be implemented in a while, I thought that at least django-rq could use runserver's auto-reloading feature.

I looked at how runserver does autoreloading and it's much simpler than I thought. Check it out yourself: https://github.com/django/django/blob/master/django/core/management/commands/runserver.py#L98-L105

Of course this may cause issues with workers being stopped mid-job or some problems across various django versions, but we should at least research this a bit more 😃 

I can prepare a PR for this.

megapctr:
Closes #262 

gif

To Do:
- [x] Docs
- [ ] Tests, if possible
- [x] Show a warning when using `--autoreload` with `DEBUG = False`?

LMK if anything else is missing!

### DIFF
--- a/django_rq/management/commands/rqworker.py
+++ b/django_rq/management/commands/rqworker.py
@@ -105,7 +105,8 @@ class Command(BaseCommand):
                 default_worker_ttl=options['worker_ttl']
             )
             if options['use_reloader']:
-                w._install_signal_handlers = lambda s: None
+                print('Reloading worker')
+                w._install_signal_handlers = lambda: None
                 w.register_death()
 
             # Call use_connection to push the redis connection into LocalStack
--- a/django_rq/management/commands/rqworker.py
+++ b/django_rq/management/commands/rqworker.py
@@ -78,7 +78,7 @@ class Command(BaseCommand):
         if options['use_reloader']:
             autoreload.main(self.inner_handle, args, options)
         else:
-            self.inner_handlel(args, **options)
+            self.inner_handle(args, **options)
 
     def inner_handle(self, *args, **options):
         # If an exception was silenced in ManagementUtility.execute in order
--- a/django_rq/management/commands/rqworker.py
+++ b/django_rq/management/commands/rqworker.py
@@ -6,6 +6,7 @@ import sys
 
 from django.core.management.base import BaseCommand
 from django.db import connections
+from django.utils import autoreload
 from django.utils.version import get_version
 
 from django_rq.queues import get_queues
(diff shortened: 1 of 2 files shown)


megapctr:
@selwin I'm having trouble figuring out what causes the tests to break. Could you try to give me a hint?

django_rq/management/commands/rqworker.py
```diff
@@ -66,12 +68,26 @@ def add_arguments(self, parser):
...
+        parser.add_argument('--autoreload', action='store_true', dest='use_reloader',
+                            help='Auto-reloads the worker code.')
 
         if LooseVersion(get_version()) >= LooseVersion('1.10'):
             parser.add_argument('args', nargs='*', type=str,
                                 help='The queues to work on, separated by space')
 
     def handle(self, *args, **options):
+        if options['use_reloader']:
+            if not settings.DEBUG:
```
kingbuzzman: i would put this block above, on line 79. you should be warning always if you’re not on debug... no?

django_rq/management/commands/rqworker.py
```diff
@@ -88,6 +104,10 @@ def handle(self, *args, **options):
                 exception_handlers=get_exception_handlers() or None,
                 default_worker_ttl=options['worker_ttl']
             )
+            if options['use_reloader']:
+                print('Reloading worker')
```
kingbuzzman: can you not use stdout? can you use logging? this trips up people like when when doing bash scripts.

django_rq/management/commands/rqworker.py
megapctr: Sorry, I don't understand. Autoreload should be used in development only. Did you want to merge these two `if`s together? I can't do that because there are more things going on than just showing the warning. Please, explain

kingbuzzman: Oh i see.. i completely read that wrong -- i apologize. You're correct. Nothing needs to happen here.

selwin:
@megapctr sorry for the late response. Looking at the test errors, I *think* the changes you made changed the code path such that `QUEUES` config is accessed before it's populated.

megapctr:
@selwin I am very, very sorry but I can't find the time to work on this PR anymore. And when I did some work, I noticed that sometimes autoreloading killed the worker in the middle of a job, which may or may not be a deal breaker for this feature.

PR #269, opened 2017-12-14
Fix RST formatting errors in README.rst

Flimm:
The PyPI page for django-rq looks really ugly, it is rendering its content as plain text, instead of accepting the RST formatting: https://pypi.python.org/pypi/django-rq

This pull request fixes the RST errors in the README.rst file. It also adds a test to make sure this gets checked before being published to PyPI.


selwin:
Thanks!

PR #270, opened 2017-12-22
More consistent usage of queue class / worker class overriding

skirsdeda:
What has been changed:
- Added `RQ.WORKER_CLASS` setting to complement `RQ.QUEUE_CLASS`;
- Changed rqworker management command to use defaults via `get_worker_class`/`get_queue_class` instead of defining it's own;
- in queues.py `get_queue` and `get_queue_by_index` now have `**kwargs` to allow passing any additional kwargs to queue class constructor;
- in worker.py add `get_worker_class` and use it in other functions;
- isort in all changed files.

### DIFF
--- a/django_rq/jobs.py
+++ b/django_rq/jobs.py
@@ -1,0 +1,23 @@
+from rq.job import Job
+from rq.utils import import_attribute
+
+from django.conf import settings
+from django.utils import six
+
+
+def get_job_class(job_class=None):
+    """
+    Return job class from RQ settings, otherwise return Job.
+    If `job_class` is not None, it is used as an override (can be
+    python import path as string).
+    """
+    RQ = getattr(settings, 'RQ', {})
+
+    if job_class is None:
+        job_class = Job
+        if 'JOB_CLASS' in RQ:
+            job_class = RQ.get('JOB_CLASS')
+
+    if isinstance(job_class, six.string_types):
+        job_class = import_attribute(job_class)
+    return job_class
--- a/django_rq/jobs.py
+++ b/django_rq/jobs.py
@@ -14,9 +14,7 @@ def get_job_class(job_class=None):
     RQ = getattr(settings, 'RQ', {})
 
     if job_class is None:
-        job_class = Job
-        if 'JOB_CLASS' in RQ:
-            job_class = RQ.get('JOB_CLASS')
+        job_class = RQ.get('JOB_CLASS', Job)
 
     if isinstance(job_class, six.string_types):
         job_class = import_attribute(job_class)
--- a/django_rq/queues.py
+++ b/django_rq/queues.py
@@ -1,11 +1,10 @@
+import redis
+from rq.queue import FailedQueue, Queue
+from rq.utils import import_attribute
+
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 from django.utils import six
-
-import redis
-from rq.utils import import_attribute
-from rq.queue import FailedQueue, Queue
-
 from django_rq import thread_queue
 
 
@@ -124,7 +123,7 @@ def get_connection_by_index(index):
 
 
 def get_queue(name='default', default_timeout=None, async=None,
-              autocommit=None, queue_class=None):
+              autocommit=None, queue_class=None, **kwargs):
     """
     Returns an rq Queue using parameters defined in ``RQ_QUEUES``
     """
@@ -140,7 +139,7 @@ def get_queue(name='default', default_timeout=None, async=None,
         queue_class = get_queue_class(QUEUES[name])
     return queue_class(name, default_timeout=default_timeout,
                        connection=get_connection(name), async=async,
-                       autocommit=autocommit)
+                       autocommit=autocommit, **kwargs)
 
 
 def get_queue_by_index(index):
@@ -181,11 +180,10 @@ def get_queues(*queue_names, **kwargs):
     All instances must use the same Redis connection.
     """
     from .settings import QUEUES
-    autocommit = kwargs.get('autocommit', None)
-    queue_class = kwargs.get('queue_class', DjangoRQ)
+
     if len(queue_names) == 0:
         # Return "default" queue if no queue name is specified
-        return [get_queue(autocommit=autocommit)]
+        return [get_queue(**kwargs)]
     if len(queue_names) > 1:
         queue_params = QUEUES[queue_names[0]]
         connection_params = filter_connection_params(queue_params)
@@ -195,7 +193,7 @@ def get_queues(*queue_names, **kwargs):
                     'Queues must have the same redis connection.'
                     '"{0}" and "{1}" have '
                     'different connections'.format(name, queue_names[0]))
-    return [get_queue(name, autocommit=autocommit, queue_class=queue_class) for name in queue_names]
+    return [get_queue(name, **kwargs) for name in queue_names]
 
 
 def enqueue(func, *args, **kwargs):
--- a/django_rq/queues.py
+++ b/django_rq/queues.py
@@ -25,16 +25,17 @@ def get_commit_mode():
     return RQ.get('AUTOCOMMIT', True)
 
 
-def get_queue_class(config):
+def get_queue_class(config={}, queue_class=None):
     """
     Return queue class from config or from RQ settings, otherwise return DjangoRQ
     """
     RQ = getattr(settings, 'RQ', {})
-    queue_class = DjangoRQ
-    if 'QUEUE_CLASS' in config:
-        queue_class = config.get('QUEUE_CLASS')
-    elif 'QUEUE_CLASS' in RQ:
-        queue_class = RQ.get('QUEUE_CLASS')
+    if queue_class is None:
+        queue_class = DjangoRQ
+        if 'QUEUE_CLASS' in config:
+            queue_class = config.get('QUEUE_CLASS')
+        elif 'QUEUE_CLASS' in RQ:
+            queue_class = RQ.get('QUEUE_CLASS')
 
     if isinstance(queue_class, six.string_types):
         queue_class = import_attribute(queue_class)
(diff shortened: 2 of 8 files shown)


skirsdeda:
One thing to think about: now multiple queue classes can be configured for different queues, however when running worker on such queues, worker needs one ``queue_class`` to work with. As I see it, this shouldn't be allowed. So I think rqworker should check queue classes of all requested queues and if they are not the same, raise ``ImproperlyConfiguredError``.

skirsdeda:
Another note: I had to ditch ``RQ`` setting proxy from ``django_rq.settings`` because it doesn't work with ``override_settings`` decorator (proxied setting is evaluated before ``override_settings`` changes it). This is not a big issue, but it would much more consistent if local settings (``django_rq.settings``) were used everywhere. Then default values (like ``{}`` for ``RQ`` setting) could be defined in one place also. And to alleviate ``override_settings`` problem, descriptors should be used for setting proxying (like django-appconf does).

skirsdeda:
* Added queue class consistency check in ``get_queues``. This seems like a proper place to do that (together with connection parameters consistency check).
* Added possibility to override job class in various places. This should be the last piece of the configuration puzzle.

So what remains is some additional tests and documentation. But I would still like to find some elegant solution for settings handling (which wouldn't break ``override_settings``).

skirsdeda:
I think this is finished. Would be great if someone could take a look at it.

django_rq/queues.py
```diff
@@ -24,16 +25,17 @@ def get_commit_mode():
     return RQ.get('AUTOCOMMIT', True)
 
 
-def get_queue_class(config):
+def get_queue_class(config={}, queue_class=None):
```
selwin: Could you please change the default argument to `None`? Using a mutable object (`{}`) as default argument often introduce hard to find bugs.

django_rq/queues.py
```diff
@@ -181,21 +185,36 @@ def get_queues(*queue_names, **kwargs):
...
+        # or one queue with specified name
+        return [get_queue(*queue_names, **kwargs)]
+
+    # will return more than one queue
+    # import job class only once for all queues
+    kwargs['job_class'] = get_job_class(kwargs.pop('job_class', None))
+
+    queue_params = QUEUES[queue_names[0]]
+    connection_params = filter_connection_params(queue_params)
+    ret = [get_queue(queue_names[0], **kwargs)]
```
selwin: Rename this variable to `first_queue`?

django_rq/queues.py
```diff
@@ -181,21 +185,36 @@ def get_queues(*queue_names, **kwargs):
...
+
+    # do consistency checks while building return list
+    for name in queue_names[1:]:
+        queue = get_queue(name, **kwargs)
+        if queue.__class__ is not ret[0].__class__:
```
selwin: I think we should use `isinstance` here.

django_rq/workers.py
```diff
@@ -12,28 +15,55 @@ def get_exception_handlers():
...
     """
     Returns a RQ worker for all queues or specified ones.
     """
-    queues = get_queues(*queue_names)
-    name = kwargs.get('name')
-    return Worker(queues,
-                  connection=queues[0].connection,
-                  exception_handlers=get_exception_handlers() or None,
-                  name=name)
+    job_class = get_job_class(kwargs.pop('job_class', None))
```
selwin: Can we use `kwargs.get('job_class')` instead? This already returns `None` by default and it doesn't mutate `kwargs` so downstream function calls can access `job_class` if needed.

django_rq/workers.py
```diff
@@ -12,28 +15,55 @@ def get_exception_handlers():
...
+    queue_class = kwargs.pop('queue_class', None)
```
selwin: We can change this to `queue_class = get_queue_class(kwargs.get('queue_class'))`. This way we'll be passing the queue class itself downstream so it doesn't need to be repeatedly imported.

django_rq/queues.py
skirsdeda: Well, that was the first thing I tried, but then `isinstance` matches all base classes. So if the first queue class is `DjangoRQ` and all others extend that, it will go through with no error.

django_rq/queues.py
skirsdeda: Agreed.

django_rq/workers.py
skirsdeda: But all downstream function calls already get `job_class` (imported here) which will never be `None`. Also, passing `**kwargs` to `worker_class` with `job_class` will probably error out with `TypeError` (got multiple values for keyword argument job_class).

django_rq/workers.py
skirsdeda: It's a bit more complicated due to possibility to configure different queue classes for each queue. So in case such configuration is used and `queue_class` kwarg is `None` here, actual queue classes would be default queue class.

django_rq/queues.py
skirsdeda: Sure.

django_rq/queues.py
```diff
@@ -197,12 +197,12 @@ def get_queues(*queue_names, **kwargs):
...
     queue_params = QUEUES[queue_names[0]]
     connection_params = filter_connection_params(queue_params)
-    ret = [get_queue(queue_names[0], **kwargs)]
+    queues = [get_queue(queue_names[0], **kwargs)]
 
     # do consistency checks while building return list
     for name in queue_names[1:]:
         queue = get_queue(name, **kwargs)
-        if queue.__class__ is not ret[0].__class__:
+        if queue.__class__ is not queues[0].__class__:
```
selwin: Can we change this to use `isinstance`? I try to avoid using `__` properties in my code base as much as possible. `isinstance` also takes class inheritance into account.

django_rq/jobs.py
```diff
@@ -0,0 +1,23 @@
...
+def get_job_class(job_class=None):
+    """
+    Return job class from RQ settings, otherwise return Job.
+    If `job_class` is not None, it is used as an override (can be
+    python import path as string).
+    """
+    RQ = getattr(settings, 'RQ', {})
+
+    if job_class is None:
+        job_class = Job
```
selwin: We can simplify these 3 lines to `job_class = RQ.get('JOB_CLASS', Job)`

django_rq/queues.py
```diff
@@ -24,16 +25,17 @@ def get_commit_mode():
...
     """
     RQ = getattr(settings, 'RQ', {})
-    queue_class = DjangoRQ
-    if 'QUEUE_CLASS' in config:
-        queue_class = config.get('QUEUE_CLASS')
-    elif 'QUEUE_CLASS' in RQ:
-        queue_class = RQ.get('QUEUE_CLASS')
+    if queue_class is None:
+        queue_class = DjangoRQ
+        if config is not None and 'QUEUE_CLASS' in config:
```
selwin:
```
if config:
    queue_class = config.get('QUEUE_CLASS', queue_class)
```

django_rq/queues.py
```diff
@@ -24,16 +25,17 @@ def get_commit_mode():
     return RQ.get('AUTOCOMMIT', True)
 
 
-def get_queue_class(config):
+def get_queue_class(config=None, queue_class=None):
     """
     Return queue class from config or from RQ settings, otherwise return DjangoRQ
     """
```
selwin: Could you please add a docstring to clarify the priorities? If I'm not mistaken, its:
1. From `queue_class`
2. From `config`
3. From `RQ` defined in settings.py

django_rq/queues.py
```diff
@@ -181,21 +185,36 @@ def get_queues(*queue_names, **kwargs):
...
+    kwargs['job_class'] = get_job_class(kwargs.pop('job_class', None))
+
+    queue_params = QUEUES[queue_names[0]]
+    connection_params = filter_connection_params(queue_params)
+    queues = [get_queue(queue_names[0], **kwargs)]
+
+    # do consistency checks while building return list
+    for name in queue_names[1:]:
+        queue = get_queue(name, **kwargs)
+        if queue.__class__ is not queues[0].__class__:
```
selwin: Please use `isinstance`

django_rq/workers.py
```diff
@@ -12,28 +15,55 @@ def get_exception_handlers():
...
+    queues = get_queues(*queue_names, **{'job_class': job_class,
```
selwin: `queues = get_queues(*queue_names, job_class=job_class, queue_class-= queue_class)`

selwin:
@skirsdeda could you also please merge the master branch into your branch? I have enabled codecov in master branch so we can check whether test coverage is still good.

django_rq/queues.py
skirsdeda: As I said, `isinstance` is exactly what we **don't** want here. `get_queues` is mostly used for worker initialization and worker must get one queue class. If we want to run worker for two queues where one is of `DjangoRQ` class and another is some `CustomQueue` extending `DjangoRQ` we cannot guarantee that `CustomQueue` has the same implementation for functionality used by worker class.

Regarding `__class__`, I can change that to `type(queue)`. Would that be better?

django_rq/jobs.py
skirsdeda: Yes.

django_rq/workers.py
skirsdeda: That wouldn't work in Python 2.7, see [PEP 3102](https://www.python.org/dev/peps/pep-3102/) which was first implemented in Python 3.

skirsdeda:
Did the rebase and requested changes.

django_rq/queues.py
skirsdeda: As I said, `isinstance` is exactly what we **don't** want here. `get_queues` is mostly used for worker initialization and worker must get one queue class. If we want to run worker for two queues where one is of `DjangoRQ` class and another is some `CustomQueue` extending `DjangoRQ` we cannot guarantee that `CustomQueue` has the same implementation for functionality used by worker class.

Regarding `__class__`, I can change that to `type(queue)`. Would that be better?

skirsdeda:
@selwin any news on this?

django_rq/queues.py
selwin: @skirsdeda you're right, I think :).

Yes, please change this to `type(queue)`

skirsdeda:
@selwin, done. Please rerun travis job, it has failed for no good reason.

selwin:
Thanks!

PR #274, opened 2018-02-19
Ability to load SENTRY_DSN for rqworker from Django settings

wormer:
I really needed this feature since I'm deploying with dokku and I don't want to store DSN in ini files.
I use environment variables and load them into settings instead. So getting DSN from settings is essential to me.

This is how it's done in https://github.com/rq/rq/blob/master/rq/cli/cli.py#L192


selwin:
Could you also document this in README?

PR #275, opened 2018-02-20
Add localization activation ability

derfenix:
Add ability to use default localization in tasks without manual activation for each task.

### DIFF
--- a/django_rq/settings.py
+++ b/django_rq/settings.py
@@ -25,3 +25,8 @@ EXCEPTION_HANDLERS = getattr(settings, 'RQ_EXCEPTION_HANDLERS', [])
 
 # Token for querying statistics
 API_TOKEN = getattr(settings, 'RQ_API_TOKEN', '')
+
+# Activate locale for each task
+RQ_USE_LOCALIZATION = getattr(settings, 'RQ_USE_LOCALIZATION', False)
+# Use specified language code to activate locale
+RQ_LANGUAGE_CODE = getattr(settings, 'RQ_LANGUAGE_CODE', settings.LANGUAGE_CODE)
--- a/django_rq/workers.py
+++ b/django_rq/workers.py
@@ -1,3 +1,4 @@
+from django.utils.translation import activate
 from rq import Worker
 from rq.utils import import_attribute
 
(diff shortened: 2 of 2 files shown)


django_rq/workers.py
```diff
@@ -38,6 +39,11 @@ def get_worker_class(worker_class=None):
     return worker_class
 
 
+def activate_localization():
+    if getattr(settings, 'RQ_USE_LOCALIZATION', False):
```
selwin: Could you please move this variable to DjangoRQ's `settings.py`?

derfenix: Done

selwin: We can simplify this to `if settings.RQ_USE_LOCALIZATION:`

django_rq/settings.py
```diff
@@ -25,3 +25,8 @@
 
 # Token for querying statistics
 API_TOKEN = getattr(settings, 'RQ_API_TOKEN', '')
+
+# Activate locale for each task
+RQ_USE_LOCALIZATION = getattr(settings, 'RQ_USE_LOCALIZATION', False)
```
selwin: Should we rename this to `RQ_USE_L10N` to keep it consistent with Django's naming scheme? `RQ_LANGUAGE_CODE` defaults to Django's `LANGUAGE_CODE` setting, should we also change this to default to Django's `USE_L10N` setting?

django_rq/workers.py
```diff
@@ -49,6 +55,7 @@ def get_worker(*queue_names, **kwargs):
     # normalize queue_class to what get_queues returns
     queue_class = queues[0].__class__
     worker_class = get_worker_class(kwargs.pop('worker_class', None))
+    activate_localization()
```
selwin: I think we can simply do this (and delete `activate_localization()`):
```python
if settings.RQ_USE_LOCALIZATION:
    activate(settings. RQ_LANGUAGE_CODE)
```

selwin:
@derfenix reminder to update this PR

selwin:
If you're still interested in getting this in, please fix the failing tests and document this feature in `README`

derfenix:
I hope I will find time in the closest week. Thanks for patience

сб, 30 мар. 2019 г. в 05:42, Selwin Ong <notifications@github.com>:

PR #276, opened 2018-02-27
Adding scheduled jobs to UI

quantumlink:
Adds Scheduled Jobs
image

and associated page to show scheduled jobs

Most of the work was already done in: https://github.com/rq/django-rq/pull/162 so thank you to @sburns

### DIFF
--- a/django_rq/queues.py
+++ b/django_rq/queues.py
@@ -255,13 +255,15 @@ instance instead of a ``Queue`` instance.
 try:
     from rq_scheduler import Scheduler
 
-    def get_scheduler(name='default', interval=60):
+    def get_scheduler(name='default', interval=60, connection=None):
         """
         Returns an RQ Scheduler instance using parameters defined in
         ``RQ_QUEUES``
         """
+        if not connection:
+            connection = get_connection(name)
         return Scheduler(name, interval=interval,
-                         connection=get_connection(name))
+                         connection=connection)
 except ImportError:
     def get_scheduler(*args, **kwargs):
         raise ImproperlyConfigured('rq_scheduler not installed')
--- a/django_rq/queues.py
+++ b/django_rq/queues.py
@@ -255,7 +255,7 @@ instance instead of a ``Queue`` instance.
 try:
     from rq_scheduler import Scheduler
 
-    def get_scheduler(name='default', interval=60, connection=None):
+    def get_scheduler(connection, name='default', interval=60):
         """
         Returns an RQ Scheduler instance using parameters defined in
         ``RQ_QUEUES``
--- a/django_rq/templates/django_rq/stats.html
+++ b/django_rq/templates/django_rq/stats.html
@@ -28,9 +28,6 @@
                     <th>Active Jobs</th>
                     <th>Deferred Jobs</th>
                     <th>Finished Jobs</th>
-                    {% if display_scheduled_jobs %}
-                        <th>Scheduled Jobs</th>
-                    {% endif %}
                     <th>Workers</th>
                     <th>Host</th>
                     <th>Port</th>
@@ -74,17 +71,6 @@
                                 -
                             {% endif %}
                         </th>
-                        {% if display_scheduled_jobs %}
-                            <td>
-                                {% if queue.name != 'failed' %}
-                                    <a href = "{% url 'rq_scheduled_jobs' queue.index %}">
-                                        {{ queue.scheduled_jobs }}
-                                    </a>
-                                {% else %}
-                                    -
-                                {% endif %}
-                            </td>
-                        {% endif %}
                         <th><a href = "{% url 'rq_workers' queue.index %}">
                                 {{ queue.workers }}
                             </a>
--- a/django_rq/templates/django_rq/stats.html
+++ b/django_rq/templates/django_rq/stats.html
@@ -28,6 +28,9 @@
                     <th>Active Jobs</th>
                     <th>Deferred Jobs</th>
                     <th>Finished Jobs</th>
+                    {% if display_scheduled_jobs %}
+                        <th>Scheduled Jobs</th>
+                    {% endif %}
                     <th>Workers</th>
                     <th>Host</th>
                     <th>Port</th>
@@ -71,6 +74,17 @@
                                 -
                             {% endif %}
                         </th>
+                        {% if display_scheduled_jobs %}
+                            <td>
+                                {% if queue.name != 'failed' %}
+                                    <a href = "{% url 'rq_scheduled_jobs' queue.index %}">
+                                        {{ queue.scheduled_jobs }}
+                                    </a>
+                                {% else %}
+                                    -
+                                {% endif %}
+                            </td>
+                        {% endif %}
                         <th><a href = "{% url 'rq_workers' queue.index %}">
                                 {{ queue.workers }}
                             </a>
--- a/django_rq/tests/tests.py
+++ b/django_rq/tests/tests.py
@@ -1,5 +1,6 @@
 import time
 import uuid
+from datetime import datetime
 from unittest import skipIf
 
 from django.contrib.auth.models import User
@@ -583,6 +584,24 @@ class ViewTest(TestCase):
             reverse('rq_deferred_jobs', args=[queue_index])
         )
         self.assertEqual(response.context['jobs'], [job])
+
+    @skipIf(RQ_SCHEDULER_INSTALLED is False, 'RQ Scheduler not installed')
+    def test_scheduled_jobs(self):
+        queue_index = get_queue_index('django_rq_test')
+
+        scheduler = get_scheduler('django_rq_test')
+        job = scheduler.enqueue_at(
+            datetime(2099, 1, 1),
+            divide,
+            [1, 2],
+        )
+
+        response = self.client.get(
+            reverse('rq_scheduled_jobs', args=[queue_index])
+        )
+        self.assertEqual(response.context['jobs'], [job])
+        job.delete()
+        scheduler.cancel(job)
 
     def test_get_all_workers(self):
         worker1 = get_worker()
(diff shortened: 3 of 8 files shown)


django_rq/views.py
```diff
@@ -15,10 +15,17 @@
                          StartedJobRegistry)
 from rq.worker import Worker
 
-from .queues import get_queue_by_index
+from .queues import get_queue_by_index, get_scheduler
 from .settings import API_TOKEN
 from .utils import get_statistics
 
+try:
```
selwin: Can we move `RQ_SCHEDULER_INSTALLED` to DjangoRQ's `settings.py`?

selwin:
Could you please modify `travis.yml` to install `rq-scheduler` prior to running the tests?

quantumlink:
Let me know if anything is off.

django_rq/utils.py
```diff
@@ -64,5 +64,10 @@ def get_statistics():
             queue_data['started_jobs'] = len(started_job_registry)
             queue_data['deferred_jobs'] = len(deferred_job_registry)
 
+            if RQ_SCHEDULER_INSTALLED:
+                scheduler = get_scheduler(queue.name)
+                queue_data['scheduled_jobs'] = len(
```
selwin: This part fetches all scheduled jobs and compares `job.origin` with `queue.name`, which could be very expensive to run (and repeats it for each queue). Let's just add a `Scheduled Jobs` row for each connection and display the number of jobs in each scheduler, similar to `FailedQueue`.

This is a stop gap measure until we build out the proper infrastructure so that scheduler can keep track of the number of scheduled jobs on a per queue basis.

django_rq/views.py
```diff
@@ -223,6 +223,43 @@ def deferred_jobs(request, queue_index):
...
+@staff_member_required
+def scheduled_jobs(request, queue_index):
+    if not RQ_SCHEDULER_INSTALLED:
+        return redirect('rq_jobs', queue_index)
+
+    queue_index = int(queue_index)
+    queue = get_queue_by_index(queue_index)
+
+    scheduler = get_scheduler(queue.name)
+    all_jobs = [job for job in scheduler.get_jobs()
```
selwin: This should just display all scheduled jobs (see [comment](https://github.com/rq/django-rq/pull/276/files#r173675281)).

quantumlink:
Done and done.

django_rq/utils.py
```diff
@@ -64,5 +64,9 @@ def get_statistics():
             queue_data['started_jobs'] = len(started_job_registry)
             queue_data['deferred_jobs'] = len(deferred_job_registry)
 
+            if RQ_SCHEDULER_INSTALLED:
+                scheduler = get_scheduler(queue.name)
+                queue_data['scheduled_jobs'] = len(scheduler.get_jobs())
```
selwin: `scheduler.count()` is a lot more efficient if we only need to know the number of scheduled jobs since we don't need to fetch all the jobs from Redis.

django_rq/views.py
```diff
@@ -223,6 +223,42 @@ def deferred_jobs(request, queue_index):
...
+        return redirect('rq_jobs', queue_index)
+
+    queue_index = int(queue_index)
+    queue = get_queue_by_index(queue_index)
+
+    scheduler = get_scheduler(queue.name)
+    all_jobs = scheduler.get_jobs()
+
+    items_per_page = 100
+    num_jobs = len(all_jobs)
```
selwin: Instead of fetching all jobs and paginating them in Python, we should use `scheduler.get_jobs(offset=foo, length=bar)`

django_rq/templates/django_rq/stats.html
```diff
@@ -28,6 +28,9 @@
                     <th>Active Jobs</th>
                     <th>Deferred Jobs</th>
                     <th>Finished Jobs</th>
+                    {% if display_scheduled_jobs %}
```
selwin: Since the scheduled aren't defined on a per queue basis, we should create a separate section to display the schedulers, one scheduler per connection, similar to `FailedQueue`.

quantumlink: Aren't Scheduled defined per queue? It defaults, but it can be specified. Am I misunderstanding something?

selwin: At the moment, `Scheduler` works like `FailedQueue` in that all jobs are stored in a single Redis key for each DB.

So doing this will store jobs in the same Redis key:
```python
scheduler = Scheduler('foo', connection=connection)
scheduler.schedule(job, ...)
scheduler = Scheduler('bar', connection=connection)
scheduler.schedule(job, ...)
```
Since jobs stored in `Scheduler` or `FailedQueue` don't necessarily belong in the same queue, DjangoRQ doesn't show `FailedQueue` as a column for each queue. `Scheduler` should be treated similarly.

django_rq/utils.py
```diff
@@ -48,6 +48,14 @@ def get_statistics():
             queue_data['started_jobs'] = '-'
             queue_data['deferred_jobs'] = '-'
 
+        elif RQ_SCHEDULER_INSTALLED and queue.name == 'scheduled':
+            scheduler = get_scheduler()
```
selwin: You need to call `get_scheduler` using the right connection here.

django_rq/views.py
```diff
@@ -40,15 +40,22 @@ def jobs(request, queue_index):
...
     queue = get_queue_by_index(queue_index)
 
-    items_per_page = 100
-    num_jobs = queue.count
     page = int(request.GET.get('page', 1))
+    items_per_page = 100
+    offset = items_per_page * (page - 1)
+
+    if RQ_SCHEDULER_INSTALLED and queue.name == 'scheduled':
+        scheduler = get_scheduler()
```
selwin: You need to call `get_scheduler()` using the right connection kwargs.

django_rq/queues.py
```diff
@@ -255,13 +255,15 @@ def get_unique_connection_configs(config=None):
 try:
     from rq_scheduler import Scheduler
 
-    def get_scheduler(name='default', interval=60):
+    def get_scheduler(name='default', interval=60, connection=None):
```
selwin: We should make the `connection` kwarg required.

selwin:
Could you please fix the test errors?

quantumlink:
I can't seem to get the tests to run so I'm going blind, but I'll try to sort it out and push an update.

django_rq/tests/tests.py
```diff
@@ -752,8 +752,9 @@ def test_get_scheduler(self):
         Ensure get_scheduler creates a scheduler instance with the right
         connection params for `test` queue.
         """
+        connection = get_connection('default')
```
selwin: You should be getting connection for `test` queue instead of default.

selwin:
Please see my comment with regards to the failed test. You also need to add a test to show that the right job count and jobs are displayed on the dashboard and job list pages.

selwin:
I plan on releasing a new version this weekend. Would be great if this feature can make the cut.

quantumlink:
I don't think I'll get the time to make the tests you wanted, I'm also confused on how they'd work. Scheduled isn't its own queue so I'm not sure how to get the right reverse_url. It's the only test that would be using the route with name `rq_jobs` and I'm not sure how to know in the test how to get the right values there. Hoping to get more time to dive into this later.

selwin:
In the case of `FailedQueue`, django-rq adds an additional `FailedQueue` for each unique Redis connection in `settings.py`. https://github.com/rq/django-rq/blob/master/django_rq/settings.py#L20

django-rq 2018-04-14 07-22-31

I think this is really confusing. If I were to implement this from scratch, I'd create a separate view `django_rq:scheduler:job_index`, with the `queue.name` as argument. This is much easier to understand.

eriktelepovsky:
Hi. I'd like to test this PR but I get the following error using master branch of @quantumlink repo:

```
Unhandled exception in thread started by <function check_errors.<locals>.wrapper at 0x115660b70>
  File "/Users/erik/env/swida/lib/python3.6/site-packages/django/utils/autoreload.py", line 225, in wrapper
    fn(*args, **kwargs)
  File "/Users/erik/env/swida/lib/python3.6/site-packages/django/core/management/commands/runserver.py", line 112, in inner_run
    autoreload.raise_last_exception()
  File "/Users/erik/env/swida/lib/python3.6/site-packages/django/utils/autoreload.py", line 248, in raise_last_exception
    raise _exception[1]
  File "/Users/erik/env/swida/lib/python3.6/site-packages/django/core/management/__init__.py", line 327, in execute
    autoreload.check_errors(django.setup)()
  File "/Users/erik/env/swida/lib/python3.6/site-packages/django/utils/autoreload.py", line 225, in wrapper
    fn(*args, **kwargs)
  File "/Users/erik/env/swida/lib/python3.6/site-packages/django/__init__.py", line 24, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/Users/erik/env/swida/lib/python3.6/site-packages/django/apps/registry.py", line 120, in populate
    app_config.ready()
  File "/Users/erik/src/django/swida/source/swida/apps.py", line 10, in ready
    self.delete_scheduled_jobs()
  File "/Users/erik/src/django/swida/source/swida/apps.py", line 14, in delete_scheduled_jobs
    scheduler = django_rq.get_scheduler('cron')
  File "/Users/erik/env/swida/src/django-rq/django_rq/queues.py", line 266, in get_scheduler
    connection=connection)
  File "/Users/erik/env/swida/lib/python3.6/site-packages/rq_scheduler/scheduler.py", line 29, in __init__
    self.connection = resolve_connection(connection)
  File "/Users/erik/env/swida/lib/python3.6/site-packages/rq/connections.py", line 66, in resolve_connection
    return patch_connection(connection)
  File "/Users/erik/env/swida/lib/python3.6/site-packages/rq/compat/connections.py", line 35, in patch_connection
    connection._hset = partial(_hset, connection)
AttributeError: 'str' object has no attribute '_hset'
```

Any ideas?

peteretep:
Has there been any progress made on this PR?

selwin:
Fixed in https://github.com/rq/django-rq/pull/400

PR #281, opened 2018-03-23
Fixed supporting for Django 2.0 and greater

koddr:
* Add fix for supporting Django `2.0` and greater
* Add fallback for Django `1.11` and older

### DIFF
--- a/django_rq/urls.py
+++ b/django_rq/urls.py
@@ -1,34 +1,73 @@
-from django.conf.urls import url
-
+import django
 from django_rq import views
 
-urlpatterns = [
-    url(r'^$',
-        views.stats, name='rq_home'),
-    url(r'^stats.json/(?P<token>[\w]+)?/?$',
-        views.stats_json, name='rq_home_json'),
-    url(r'^queues/(?P<queue_index>[\d]+)/$',
-        views.jobs, name='rq_jobs'),
-    url(r'^workers/(?P<queue_index>[\d]+)/$',
-        views.workers, name='rq_workers'),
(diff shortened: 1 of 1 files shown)


django_rq/urls.py
```diff
@@ -1,34 +1,73 @@
...
+        re_path(r'^queues/(?P<queue_index>[\d]+)/(?P<job_id>[-\w]+)/delete/$',
+                views.delete_job, name='rq_delete_job'),
+        re_path(r'^queues/actions/(?P<queue_index>[\d]+)/$',
+                views.actions, name='rq_actions'),
+        re_path(r'^queues/(?P<queue_index>[\d]+)/(?P<job_id>[-\w]+)/requeue/$',
+                views.requeue_job_view, name='rq_requeue_job'),
+    ]
+
+# Fallback for Django 1.11 and lower
+if django.VERSION[0] == 1:
```
selwin: We should avoid declaring all the URL paths twice. According to the docs: `The django.conf.urls.url() function from previous versions is now available as django.urls.re_path()`. So you should be able to do this:
```python
try:
    from django.urls import re_path  # Django 2.x
except ImportError:
    from django.conf.urls import url as re_path # Django 1.x

urlpatterns = [
    re_path(r'^$', views.stats, name='rq_home'),
    ...
]


```

koddr: Thanks, I fixed it.

selwin:
Thanks!

koddr:
@selwin waiting for new version!

selwin:
I'll make a new release this weekend or early next week. Hoping this https://github.com/rq/django-rq/pull/276 is ready by then.

PR #283, opened 2018-04-18
Added support for using redis sentinels.

SpeedyCoder:
The idea here is to be able to specify in settings a list of sentinels from which the connection to master will be obtained. It seems to be working fine for me, but I may have missed something. Can you please have a look @selwin ?


selwin:
I looked at this PR and it looks alright. Having said that, I actually don't have a lot of experience using Sentinel, nor do I have an easy way to test this out so I'll trust you on this one :)

PR #285, opened 2018-04-20
Removed fixed table width.

tystakartografen:
Noticed a minor styling issue in the Django RQ admin section. The table appears to be set to a fixed width of 498px, this commit aims to resolve the issue by setting the width to 100%.

**Bug:** _Width set to 498px._
screen shot 2018-04-20 at 21 47 19

**Fix:** _Width set to 100%._
screen shot 2018-04-20 at 21 47 37


selwin:
Thanks!

PR #288, opened 2018-04-25
In README.rst change the use of url() for Django >= 2.0 to path().

richardgiddings:
Change README.rst to use path() instead of url() in the urls.py section to make it more Django 2.0


DefCon-007:
Was struggling with this issue for a long time. The changes work as expected.

selwin:
Thanks!

PR #291, opened 2018-05-28
Handle unpickle error

selwin:


lifenautjoe:
Hi @selwin , do you think this has anything to do with https://github.com/rq/django-rq/issues/358 ?

Thanks!

PR #293, opened 2018-06-11
How to run the RQ Worker as a service

W7PEA:
I hope this is helpful to those who come next... Could us some additional refinement I'm sure.


selwin:
Thanks!

PR #297, opened 2018-07-17
Rename `async` to `is_async` to support Python 3.7

### LINKED ISSUE #295: "async" function argument causing SyntaxError since Python3.7
In django_rq/queues.py:

    def get_queue(name='default', default_timeout=None, async=None,
                  autocommit=None, queue_class=None, job_class=None, **kwargs):
        ...

Python3.7's release note:

    Backwards incompatible syntax changes:

        async and await are now reserved keywords.

Flimm:
Note that the new version of the dependency rq also did the same rename
for the same reason.

The variable `_async` was also renamed to `_is_async`.

Test in Python 3.7 and mark 3.7 as supported.

Fixes #295

### DIFF
--- a/django_rq/queues.py
+++ b/django_rq/queues.py
@@ -146,7 +146,7 @@ def get_queue(name='default', default_timeout=None, is_async=None,
     """
     from .settings import QUEUES
 
-    if 'async' in kwargs and kwargs['async'] is not None:
+    if kwargs.get('async') is not None:
         is_async = kwargs['async']
         warnings.warn('The `async` keyword is deprecated. Use `is_async` instead', DeprecationWarning)
 
--- a/django_rq/queues.py
+++ b/django_rq/queues.py
@@ -1,3 +1,5 @@
+import warnings
+
 import redis
 from redis.sentinel import Sentinel
 from rq.queue import FailedQueue, Queue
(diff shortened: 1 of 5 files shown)


django_rq/queues.py
```diff
@@ -137,24 +139,28 @@ def get_connection_by_index(index):
...
     Returns an rq Queue using parameters defined in ``RQ_QUEUES``
     """
     from .settings import QUEUES
 
-    # If async is provided, use it, otherwise, get it from the configuration
-    if async is None:
-        async = QUEUES[name].get('ASYNC', True)
+    if 'async' in kwargs and kwargs['async'] is not None:
+        is_async = kwargs['async']
+        warnings.warn('The `async` keyword is deprecated. Use `is_async` instead', DeprecationWarning)
```
samuelcolvin: Could you just add a test for this case.

django_rq/queues.py
```diff
@@ -137,24 +139,28 @@ def get_connection_by_index(index):
...
               autocommit=None, queue_class=None, job_class=None, **kwargs):
     """
     Returns an rq Queue using parameters defined in ``RQ_QUEUES``
     """
     from .settings import QUEUES
 
-    # If async is provided, use it, otherwise, get it from the configuration
-    if async is None:
-        async = QUEUES[name].get('ASYNC', True)
+    if 'async' in kwargs and kwargs['async'] is not None:
```
selwin: We can simplify this to `if kwargs.get('async') is not None:

selwin:
Thanks!

tomhamiltonstubber:
Can you do a release for this pls?

PR #303, opened 2018-09-19
Fix how django-rq connects to Sentinel's

### LINKED ISSUE #302: When sentinel is used it does not respect socket_timeout
Hello,

we are using rq/django-rq together with redis sentinel.
And we experience some error when one of sentinel instance went down.

**Environment**
4 instances with redis/sentinel (2 instance for each DC)

**Bad behavior**
When using django-rq with sentinel it does not respect (socket_timeout), so when one sentinel instance went down (planned maintenance, system upgrade or failure) django-rq hangsout until request timeout. 

**Detailed explanation**
Currently Sentinel is initialized like [that](https://github.com/rq/django-rq/blob/v1.2.0/django_rq/queues.py#L117-L121)
```python
        sentinel = Sentinel(config['SENTINELS'])
        return sentinel.master_for(
            service_name=config['MASTER_NAME'], redis_class=redis_cls, db=config.get('DB'),
            password=config.get('PASSWORD'), socket_timeout=config.get('SOCKET_TIMEOUT')
        )
```

When new connection is made redis-py calls [this method](https://github.com/andymccurdy/redis-py/blob/2.10.6/redis/connection.py#L978-L983)
```python
    def make_connection(self):
        "Create a new connection"
        if self._created_connections >= self.max_connections:
            raise ConnectionError("Too many connections")
        self._created_connections += 1
        return self.connection_class(**self.connection_kwargs)
```

The most important line is 
`self.connection_class(**self.connection_kwargs)`

Since Sentinel was initialized without any `kwargs` and inside `Sentinel.__init__` method it creates multiple instances of `StrictRedis` instances (without knowledge about our `socket_timeout`)

In some cases when new connection is made `self.connection_kwargs` is empty and value for `socket_timeout` is None. Under the neath connection is made using `socket` module [like that](https://github.com/andymccurdy/redis-py/blob/2.10.6/redis/connection.py#L526-L532)
```python
                sock.settimeout(self.socket_connect_timeout)

                # connect
                sock.connect(socket_address)

                # set the socket_timeout now that we're connected
                sock.settimeout(self.socket_timeout)
``` 
from socket.settimeout [documentation](https://docs.python.org/3/library/socket.html#socket.socket.settimeout)
```
.... If None is given, the socket is put in blocking mode.
```

at this point when at least one sentinel instance is unavailable, request to (for example) `/django/django-rq/` will wait until 504 timeout (in my case)

PR with fix is almost ready.

pnuckowski:
* fixes #302 
* added support to pass additional connection kwargs to Sentinel

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -48,11 +48,14 @@ Installation
             'DEFAULT_TIMEOUT': 360,
         },
         'with-sentinel': {
-           'SENTINELS': [('localhost', 26736), ('localhost', 26737)],
-           'MASTER_NAME': 'redismaster',
-           'DB': 0,
-           'PASSWORD': 'secret',
-           'SOCKET_TIMEOUT': None,
+            'SENTINELS': [('localhost', 26736), ('localhost', 26737)],
+            'MASTER_NAME': 'redismaster',
+            'DB': 0,
+            'PASSWORD': 'secret',
+            'SOCKET_TIMEOUT': None,
+            'CONNECTION_KWARGS': {
+                'socket_connect_timeout': 0.3
+            },
         },
         'high': {
             'URL': os.getenv('REDISTOGO_URL', 'redis://localhost:6379/0'), # If you're on Heroku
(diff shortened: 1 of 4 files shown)


selwin:
Merged. Thanks!

PR #304, opened 2018-10-24
provide connection as keyword argument for get_queue method

kprikshit:
this will allow you to provide your own redis connection object instead of creating one internally. 

This is done because every time we create a queue, it ends up creating a new connection object which in turn causes too many connections in redis.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -110,12 +110,25 @@ Putting jobs in the queue
     queue.enqueue(func, foo, bar=baz)
 
 In addition to ``name`` argument, ``get_queue`` also accepts ``default_timeout``,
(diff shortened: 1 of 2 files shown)


selwin:
Thanks, this is a good addition. Could you please document this argument in README?

kprikshit:
Sure. I'll do it. 

> Thanks, this is a good addition. Could you please document this argument in README?

kprikshit:
@selwin Hey mate, I think all CI checks are failing due to Redis 3. Can you have a look and merge this accordingly ?

If this gets released, I'll make changes in my production accordingly. I am running a hack now.

selwin:
Thanks. Yeah, I'll tag a release soon.

PR #306, opened 2018-11-07
Add django rollbar integration

mangeld:


selwin:
Ideally, I think rollbar's integration should be done in RQ and Django-RQ just piggy backs the feature that's already built into RQ. Care to open a PR there, complete with docs?

mangeld:
Hi @selwin, I get your point but I think integrations to error reporters should be done in the reporters themselves like sentry and rollbar does, and the initialisation should be done by the user.

Since django-rq is an abstraction of rq I think putting that initialisation here is the right place because anyone using a standalone rq would be free of choosing whichever error reporter they wanted but with django-rq that is not possible since we abstract the configuration of rq, thus we have to support different error reporters and enabling their integrations with rq.

PR #307, opened 2018-11-08
Enqueue job view

SeiryuZ:
I added a view to enable user to force enqueue deferred job

### DIFF
--- a/django_rq/templates/django_rq/job_detail.html
+++ b/django_rq/templates/django_rq/job_detail.html
@@ -174,6 +174,12 @@
                 <input type="submit" value="Requeue" class="default" name="requeue">
             </form>
         {% endif %}
+        {% if job.is_deferred  %}
+            <form method = 'POST' action = "{% url 'rq_enqueue_job' queue_index job.id %}">
+                {% csrf_token %}
+                <input type="submit" value="Enqueue" class="default" name="Enqueue">
+            </form>
+        {% endif %}
     </div>
 
 </div>
--- a/django_rq/templates/django_rq/job_detail.html
+++ b/django_rq/templates/django_rq/job_detail.html
@@ -174,12 +174,10 @@
                 <input type="submit" value="Requeue" class="default" name="requeue">
             </form>
         {% endif %}
-        {% if job.is_deferred  %}
-            <form method = 'POST' action = "{% url 'rq_enqueue_job' queue_index job.id %}">
-                {% csrf_token %}
-                <input type="submit" value="Enqueue" class="default" name="Enqueue">
-            </form>
-        {% endif %}
+        <form method = 'POST' action = "{% url 'rq_enqueue_job' queue_index job.id %}">
+            {% csrf_token %}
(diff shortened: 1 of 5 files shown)


django_rq/templates/django_rq/job_detail.html
```diff
@@ -174,6 +174,12 @@
                 <input type="submit" value="Requeue" class="default" name="requeue">
             </form>
         {% endif %}
+        {% if job.is_deferred  %}
+            <form method = 'POST' action = "{% url 'rq_enqueue_job' queue_index job.id %}">
```
selwin: I think we should just show this button regardless of the job’s status. Sometimes I have the need to requeue finished jobs so this is a useful option to have.

django_rq/views.py
```diff
@@ -368,3 +370,30 @@ def actions(request, queue_index):
...
+def enqueue_job_view(request, queue_index, job_id):
+    """ Enqueue deferred jobs
+    """
+    queue_index = int(queue_index)
+    queue = get_queue_by_index(queue_index)
+    registry = DeferredJobRegistry(queue.name, queue.connection)
+    job = Job.fetch(job_id, connection=queue.connection)
+
+    if request.method == 'POST':
+        job.enqueued_at = utcnow()
```
selwin: You should use `queue.enqueue_job()`

django_rq/views.py
```diff
@@ -368,3 +370,30 @@ def actions(request, queue_index):
                 messages.info(request, 'You have successfully requeued %d  jobs!' % len(job_ids))
 
     return redirect('rq_jobs', queue_index)
+
+
+@staff_member_required
+def enqueue_job_view(request, queue_index, job_id):
```
selwin: Can you rename this to `enqueue_job`?

SeiryuZ: This would broke an existing naming scheme though, line 274 with `requeue_job_view` unless you want me to change that as well

selwin: `requeue_job_view` is called that because `requeue_job` is already a function imported from RQ.

django_rq/views.py
```diff
@@ -230,6 +230,7 @@ def job_detail(request, queue_index, job_id):
     queue = get_queue_by_index(queue_index)
     try:
         job = Job.fetch(job_id, connection=queue.connection)
+        job.is_deferred = (job.get_status() == JobStatus.DEFERRED)
```
selwin: We no longer need this.

django_rq/views.py
```diff
@@ -368,3 +369,34 @@ def actions(request, queue_index):
...
+@staff_member_required
+def enqueue_job(request, queue_index, job_id):
+    """ Enqueue deferred jobs
+    """
+    queue_index = int(queue_index)
+    queue = get_queue_by_index(queue_index)
+    job = Job.fetch(job_id, connection=queue.connection)
+
+    # Determine correct registry to remove job from
+    if job.get_status() == JobStatus.DEFERRED:
```
selwin: This check only needs to be done `if request.method == 'POST'`

PR #316, opened 2019-01-15
Fix: change load admin_static to load static. fix #315

### LINKED ISSUE #315: {% load admin_static %} DeprecationWarning
All the templates from django_rq use {% load admin_static %} which will be removed in djaango 3.0 in favor of {% load static %}

https://docs.djangoproject.com/en/dev/releases/2.1/#deprecated-features-2-1

Can we change it?

lorenzomorandini:


selwin:
Thanks!

Issue #319, opened 2019-01-22
Support Sentry with sentry-sdk


Bolayniuss:
Since April 2018 Sentry has release it's own package (`sentry-sdk`) to report issues. Using raven is now deprecated. With this package they have dropped the password part of the DSN. Therefore `django-rq` is not compatible with project using `sentry-sdk` as raven internally requires that the DSN is composed of a username *AND*  a password.

`sentry-sdk` has a built-in integration of RQ. According to the doc (https://docs.sentry.io/platforms/python/rq/)  we should just by-pass the registration of sentry in RQ by setting `--sentry_dsn` to an empty string.

selwin:
Support for Sentry-SDK should come in a few weeks time. I'm working on building this into RQ 1.0, which should be released this month.

selwin:
Fixed in https://github.com/rq/django-rq/pull/333/commits/324b79e0b8c34ed03fc29325acac69d62ba6cbfe

PR #320, opened 2019-01-22
add argument to disable sentry integration

Bolayniuss:
allow disabling sentry integration in the case where the project use sentry-sdk as `SENTRY_DSN` is not compatible between `sentry_sdk` and `raven` (missing secret key, a.k.a. password)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -308,10 +308,11 @@ Also you can specify ``sentry-dsn`` parameter when running rqworker:
 
 ``./manage.py rqworker --sentry-dsn=https://*****@sentry.io/222222``
 
-If your project use ``sentry-sdk``, the DSN is not compatible with the sentry integration
-used inside RQ.
 
-In that case you can disable sentry integration by specifying ``--disable-sentry-integration``.
+**Disable RQ sentry plugin**
+
+If your project use ``sentry-sdk``, the DSN is not compatible with RQ's sentry plugin (based on raven).
+In that case you have to disable the sentry plugin by setting `--sentry-dsn=""`.
 
 Configuring Logging
 -------------------
--- a/README.rst
+++ b/README.rst
@@ -308,6 +308,11 @@ Also you can specify ``sentry-dsn`` parameter when running rqworker:
 
 ``./manage.py rqworker --sentry-dsn=https://*****@sentry.io/222222``
 
+If your project use ``sentry-sdk``, the DSN is not compatible with the sentry integration
+used inside RQ.
+
+In that case you can disable sentry integration by specifying ``--disable-sentry-integration``.
+
 Configuring Logging
 -------------------
 
--- a/django_rq/management/commands/rqworker.py
+++ b/django_rq/management/commands/rqworker.py
@@ -47,8 +47,6 @@ class Command(BaseCommand):
                             help='Default worker timeout to be used')
         parser.add_argument('--sentry-dsn', action='store', default=None, dest='sentry-dsn',
                             help='Report exceptions to this Sentry DSN')
-        parser.add_argument('--disable-sentry-integration', action='store_true', dest='disable-sentry',
-                            help='Disable sentry integration. Useful if `sentry-sdk` is used')
 
         if LooseVersion(get_version()) >= LooseVersion('1.10'):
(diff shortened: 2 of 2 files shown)


Bolayniuss:
link to https://github.com/rq/django-rq/issues/319

README.rst
```diff
@@ -308,6 +308,11 @@ Also you can specify ``sentry-dsn`` parameter when running rqworker:
 
 ``./manage.py rqworker --sentry-dsn=https://*****@sentry.io/222222``
 
+If your project use ``sentry-sdk``, the DSN is not compatible with the sentry integration
+used inside RQ.
```
mbodock: I believe you could mention raven here. to make things a bit clear.

README.rst
```diff
@@ -308,6 +308,11 @@ Also you can specify ``sentry-dsn`` parameter when running rqworker:
...
+
+In that case you can disable sentry integration by specifying ``--disable-sentry-integration``.
```
mbodock: Also mention `disable sentry integration by raven` ...

Also could help others locate it easily inside the readme

Bolayniuss:
Maybe it would be better to use `--sentry_dsn=""` instead of adding an extra argument. This would better fit with what sentry propose on it's RQ integration doc (https://docs.sentry.io/platforms/python/rq/).

Actually the condition to use `SENTRY_DSN` set in the django settings is just `if sentry_dsn:`. Disabling sentry with `--sentry_dns=""` would just requires something like:

Bolayniuss:
@mbodock take a look at this modification, for me this is far better than the first solution I've proposed.

mbodock:
I agree with you. This looks very fine to me.

@selwin anything that we can do to speed this up?

I believe Travis failed to get a file from S3, probably momentary network error.

selwin:
See https://github.com/rq/rq/issues/1017 with regards to Sentry integration.

If `--sentry-sdk-dsn` flag is passed, we do this:
```
import sentry_sdk
from sentry_sdk.integrations.rq import RqIntegration

sentry_sdk.init("https://foo@sentry.io/123", integrations=[RqIntegration()])
```

When `--sentry-dsn` is passed in, raise a `DeprecationWarning` along with an approprioate message.

Bolayniuss:
I don't see any problem having different ways to deal with sentry integration in `django-rq` or `rq`.

In general I don't like that magic stuff happens in a library, based on an env variable, specially if there is no way to disable it.

I think that allowing an empty string int the `--sentry_dsn` is an easy solution to implement AND to use and is moreover backward compatible.

awbacker:
@selwin @Bolayniuss 

A flag to disable the `rq` sentry feature would be a good stopgap.  We would still have the issue that we are also installing `raven` though we don't use it.  No matter what `rq` does there will always be a conflict between the app using the version of sentry (or rollbar!) it wants, but `rq` insisting on something else.  I looked at the related issue (rq/rq#1017) as well.

`django-rq` especially, and likely `rq`, should know nothing of sentry at all.  I don't see it as a library or framework's job to bake in a error service, since conflict is ineveitable.

Another stopgap may be allowing the django-rq manage.py commands to forcibly disable that sentry feature in `rq` by starting it up differently. 

A few things make sense to me as ways to add custom exception logging:

* `rq` could use standard logger.error(exc_info=True)
  * This is already picked up by sentry w/o an integration
* `rq` could document a separate user-created `on_init.py` 
  * Rq could look for this an import it if found.  This file would not be added when using `django-rq` since in that case we expect the user to be init-ing sentry themselves if they are using the service at all.
* `rq` could have an `on_exception` handler.  Our app can pass sentry's `capture_exception(e)` func
* An internal `rq.contrib.sentry.DjangoIntegration`?
* A `rq[sentry]` option

I think Bolay and myself would be happy to help out with this kind of work a little, if the core maintainers of the two packages can settle on a path forward.

selwin:
Ok, let's go with this solution to fix the current mess. @Bolayniuss could you please add fix the conflicts so I can merge this in?

Bolayniuss:
@selwin conflict resolved

selwin:
Thanks!

PR #321, opened 2019-02-13
Logging configuration for worker and scheduler

### LINKED ISSUE #91: Autoconfigure logging for the ``rqscheduler`` management command also
The `rqworker` management command has a section that setups logging for `rq.worker` if it hasn't been configured in `settings.py`. I recommend adding a similar section to the `rqscheduler` management command, something like: 

```
logger = logging.getLogger('rq_scheduler')
if not logger.handlers:
    logger.setLevel(logging.DEBUG)
    formatter = logging.Formatter(fmt='%(asctime)s %(message)s',
                                  datefmt='%H:%M:%S')
    handler = ColorizingStreamHandler()
    handler.setFormatter(formatter)
    logger.addHandler(handler)
```

Both management commands should have a similar behavior since users (like myself) may be running `python manage.py rqscheduler` and be confused on the lack of any output (if they haven't explicitly setup loggers for `rq_scheduler`). 

Is there a reason for not configuring loggers on `rqscheduler` ?

If you want I could provide a patch - however I wanted first to start a discussion for that. 

Thanks !

### LINKED ISSUE #191: Improve logging auto-configuration
django-rq attempts to autoconfigure the `'rq.worker'` logger here:
https://github.com/ui/django-rq/blob/43b6eb344ceb7d0be9677aae544850855237eba2/django_rq/management/commands/rqworker.py#L18-L26

This pattern was copied from `rq` itself, however, the check for an existing logging configuration is incorrect and initializes the logger even if its messages would be caught by an explicitly, properly configured root logger.

https://github.com/nvie/rq/commit/fe122ac7993b972ff2230fe4987a0a23b4c43e79 made changes and slightly improved the situation, but it's still incorrect, as described in https://github.com/nvie/rq/issues/750.

This ticket applies to django-rq exactly like it does to rq. Relevant parts:

I would like to suggest the following pattern:

``` python
try:
    logger.hasHandlers
except AttributeError:
    # Compatibility shim for Python < 3.2
    has_handlers = bool(logger.handlers)
else:
    has_handlers = logger.hasHandlers()

if not has_handlers:
    # automagically configure logging
```

You could also use something more complicated than `bool(looger.handlers)` for Python 2 (but I don't care about that part, it would make sense to favor stability above correctness on Python 2 at that point).

Paulius-Maruska:
Both `rq` and `rq_shceduler` have their own version of `setup_loghandlers` function in their code. This PR updates both `rqworker` and `rqscheduler` commands to call those functions.

This closes #91, closes #191.

### DIFF
--- a/django_rq/management/commands/rqscheduler.py
+++ b/django_rq/management/commands/rqscheduler.py
@@ -1,9 +1,16 @@
 import os
 from distutils.version import LooseVersion
 
+from django.core.exceptions import ImproperlyConfigured
 from django.core.management.base import BaseCommand
 from django.utils.version import get_version
 from django_rq import get_scheduler
+
+try:
+    from rq_scheduler.utils import setup_loghandlers
+except ImportError:
+    def setup_loghandlers(*args, **kwargs):
+        raise ImproperlyConfigured('rq_scheduler not installed')
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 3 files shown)


dotpot:
Finally!!! Well done! 👍

selwin:
Thanks!

Issue #292, opened 2018-05-28
rqscheduler does not use TIMEOUT set in settings.py


nmiletic:
Hello,

It seems that rqscheduler does not use TIMEOUT set in settings.py.
when Queue instance is returned using from_queue key timeout is not set and default of 180sec is used it seems. 

Thanks
Nemanja

selwin:
Could you please open a PR for this?

simone6021:
+1 for this: really appreciated and needed.

@nmiletic, i am available to complete this PR.

selwin:
Fixed in https://github.com/rq/django-rq/pull/332

Issue #278, opened 2018-03-08
Dose rqscheduler support custom queue_class?


legshort:
It seems like `get_scheduler()` calls without import custom `queue_class` from django `setting` file.

selwin:
PR welcome :)

PR #332, opened 2019-03-29
Scheduler now respects default timeout and result_ttl

simone6021:
As per subject, solves #292.

Thanks for the awesome work!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ matrix:
       env: DJANGO=1.11.16
 
 install:
-  - pip install Django==$DJANGO times docutils pygments
+  - pip install Django==$DJANGO times docutils pygments rq-scheduler
   - python setup.py install
   - pip install codecov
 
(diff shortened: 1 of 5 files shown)


selwin:
Before I merge this in, could you please improve the test coverage by modifying https://github.com/rq/django-rq/blob/master/.travis.yml#L36 to install `rq-scheduler` prior to running tests?

simone6021:
Here you are.  

Should document the possibility to set a custom scheduler class in settings and maybe add it to the scheduler commad.

django_rq/.kdev4/django_rq.kdev4
```diff
@@ -0,0 +1,17 @@
+[Buildset]
```
selwin: This file shouldn't be committed. Could you please remove this file from the repo?

django_rq/django_rq.kdev4
```diff
@@ -0,0 +1,4 @@
+[Project]
```
selwin: This file should also be removed.

simone6021:
This should solve also #278

selwin:
Thanks!

PR #336, opened 2019-04-18
Remove redundant sentry_dsn config line

alexbarbato:
In our local environment we use the config var "SENTRY_SDK" for our actual sentry_sdk package variable in settings.

With that variable set I was unable to set --sentry-sdk="" with the environment variable still set because the line deleted would override the option passed in and thus rqworker would crash until we changed our config variable to something other than SENTRY_SDK.


selwin:
Thanks!

Issue #339, opened 2019-04-29
django-rq should declare sentry-sdk as a dependency


sconstable:
We recently updated our `requirements.txt` file and pulled in the new version of django-rq.  We recently encountered an issue where in development and CI everything worked fine but when deployed to staging everything broke.  It turns out that since our staging server uses Sentry, the `sentry_dsn` flag gets set to True and django-rq fails to import sentry-sdk, prints an error message instructing the user to install sentry-sdk, and exits.  This is definitely not expected behaviour, and could be easily fixed by declaring sentry-sdk as a dependency in `setup.py` so that other people don't get surprised by this.

selwin:
sentry-sdk is only required if you use Sentry so I don't want to add it as a dependency in `setup.py`. This behavior should be similar to previous versions of django-rq which requires `raven` to be installed if error reporting to Sentry is enabled.

sconstable:
I understand wanting to keep the codebase clean, but the way things are set up right now is such that if I don't care about enabling Sentry support for Django-RQ then my code will break if I have Sentry enabled on production servers.  

I strongly suggest reworking the [affected lines](https://github.com/rq/django-rq/blob/master/django_rq/management/commands/rqworker.py#L91) such that even if `SENTRY_DSN` is defined in `settings.py` the code will still run instead of just exiting.

selwin:
I merged in this PR that lets you run `./manage.py rqworker --sentry-dsn=""` to turn off Sentry. https://github.com/rq/django-rq/pull/336 . You can use this to disable Sentry reporting if you don't need it.

PR #340, opened 2019-04-29
Added explicit dependency on `sentry-sdk`

sconstable:
This is in relation to a submitted issue, #339


selwin:
Thanks for the PR. See my comment https://github.com/rq/django-rq/issues/339#issuecomment-489414242

PR #354, opened 2019-06-24
Bug Fix: Fix NOAUTH when using Sentinel with Password

neil-pozetroninc:
Currently when using Sentinel the password from the config is not passed through to the Sentinel class which results in a NOAUTH when connecting.

This patch passes through the same `PASSWORD` from the config through to both the Sentinel and the Master. This may not be what you want but it will solve the problem if both passwords are the same.


selwin:
Could you please fix the test errors so I can merge this in?

neil-pozetroninc:
I can, but is this the approach you want to take? There are several ways to address this bug, for example you may want another config value for the SENTINEL password instead.

Please let me know if this is your preferred approach.

selwin:
I think it would be better if the user can explicitly specify auth credentials to sentinels.

Issue #202, opened 2016-11-10
Extend url pattern for job_id?


pacahon:
Hello.

Queue `enqueue` method allows to provide custom `job_id` instead of autogenerated `uuid4`.
I believe, that colon sign `:` is very useful for redis keys (I might be wrong).

But if we look at `urls.py`:

```
url(r'^queues/(?P<queue_index>[\d]+)/(?P<job_id>[-\w]+)/$',
        views.job_detail, name='rq_job_detail'),
```

In fact, this sign is not allowed and view fails with `NoReverseMatch` exception.
Maybe we can extend default url pattern for most popular symbols? At least for `:`.

selwin:
Seems like a reasonable idea. Feel free to submit a PR for this :)

PR #359, opened 2019-08-30
Allow viewing of jobs that have colons in their IDs

takkaria:
This is a pull request for #202, which is an issue I've run into in my work (I can't view jobs which have colons in their job ids).


takkaria:
I am pretty sure that the codecov report is inaccurate.

selwin:
Thanks!

Issue #366, opened 2019-10-09
Make disabling default sentry initialization more obvious


karech:
Currently docs says:
> Disable RQ sentry plugin

> If your project use sentry-sdk, the DSN is not compatible with RQ's sentry plugin (based on raven). In that case you have to disable the sentry plugin by setting --sentry-dsn="".

This is not obvious and looks not good.

Better to allow do it via settings.

selwin:
Actually versions > 2.0 is compatible with `sentry-sdk`, so I'll just have to update the docs about disabling Sentry plugin.

karech:
@selwin Hey, the problem is that passing `--sentry-dsn=""` to disable default sentry integration is not obvious.

For example, in our project we have custom sentry-sdk setup, but django-rq tries to init it anyway if finds SENTRY_SDN in django.settings. 
And passing `--sentry-dsn=""` to django-rq worker is not looking good and obvious.

selwin:
Ah ok, in that case could you please reopen the PR and update the docs accordingly? Thanks

karech:
@selwin I can't reopen it (and this issue), so could you do it please?

Also in PR I've updated [README.rst](https://github.com/rq/django-rq/pull/367/files#diff-88b99bb28683bd5b7e3a204826ead112R321) already.
Do I need to update any other docs?

karech:
Created new issue: https://github.com/rq/django-rq/issues/372

PR #367, opened 2019-10-09
Allow disabling sentry initialization in more obvious way

karech:
for issue #366  

Done:
1. issue fix
2. tests
3. docs update


selwin:
Django-RQ >= 2.0 works with `sentry-sdk` so the docs about disabling Sentry have been removed.

PR #386, opened 2019-12-04
Honor logging level set by --verbosity

stlk:
I discovered that verbosity parameter is not honored because [worker gets called with default info parameter](https://github.com/rq/rq/blob/master/rq/worker.py#L443).


selwin:
Thanks!

PR #388, opened 2019-12-08
update trove classifiers for Python 3.8

terencehonles:


terencehonles:
@selwin we should also update the CHANGELOG for 2.2.0. I don't know if there's anything other than Python 3.8 support in that release.

terencehonles:
That looks like it covers anything. I was just looking at `git log master...v2.1.0` it also looks like there needs to be a tag pushed too.

selwin:
I already took care of the changelog and tagging of the release yesterday, right before I uploaded the package to PyPI, but I forgot to push the changes to Github. It's fixed now, could you please update the PR so I can pull in changes to the classifiers?

terencehonles:
@selwin sorry for the delay, I removed my other changes so it's just the trove classifiers now.

Issue #390, opened 2019-12-09
NoSuchJobError not handled when requeueing from failed registry


thomasmatecki:
When attempting to _requeue all_ failed jobs in a queue, [`NoSuchJobError`](https://github.com/rq/rq/blob/cf501518534429640a9a6d14e022f9329225e701/rq/exceptions.py#L6) may be raised and not handled gracefully.

It appears to occur in this loop:
https://github.com/rq/django-rq/blob/f63b984f4bc1be7df39128c98c0ccdd89db0c226/django_rq/views.py#L359

I propose we simply catch this exception and set a warning message.

selwin:
PR welcome :)

selwin:
Fixed in https://github.com/rq/django-rq/pull/392

PR #392, opened 2019-12-28
Gracefully hand NoSuchJobError when requeuing all jobs

thomasmatecki:
Address #390 

Gracefully continue if a job being requeued no longer exists. Update messaging to be a little more precise. :small_airplane:


selwin:
Thanks!

PR #400, opened 2020-01-28
Stats for scheduled jobs

Yolley:
Starting with 1.2.0 rq  supports job scheduling without additional requirements, so it would be nice to support stats for scheduled jobs.

This line https://github.com/rq/django-rq/pull/400/files#diff-51cc121b676f5b91f7ee8499cc586a1dR173 may not be the best idea, but I don't see a clearer way of adding this attribute without additional changes in rq.

### DIFF
--- a/django_rq/templates/django_rq/stats.html
+++ b/django_rq/templates/django_rq/stats.html
@@ -31,7 +31,7 @@
                     <th>Deferred Jobs</th>
                     <th>Finished Jobs</th>
                     <th>Failed Jobs</th>
-                    <th>Scheduled jobs</th>
+                    <th>Scheduled Jobs</th>
                     <th>Workers</th>
                     <th>Host</th>
                     <th>Port</th>
--- a/django_rq/templates/django_rq/stats.html
+++ b/django_rq/templates/django_rq/stats.html
@@ -31,6 +31,7 @@
                     <th>Deferred Jobs</th>
                     <th>Finished Jobs</th>
                     <th>Failed Jobs</th>
+                    <th>Scheduled jobs</th>
                     <th>Workers</th>
                     <th>Host</th>
                     <th>Port</th>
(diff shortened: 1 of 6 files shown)


django_rq/templates/django_rq/stats.html
```diff
@@ -31,6 +31,7 @@
                     <th>Deferred Jobs</th>
                     <th>Finished Jobs</th>
                     <th>Failed Jobs</th>
+                    <th>Scheduled jobs</th>
```
selwin: Nitpick: `Scheduled Jobs` (should be title case :)

django_rq/views.py
```diff
@@ -144,6 +149,47 @@ def failed_jobs(request, queue_index):
...
+
+    if num_jobs > 0:
+        last_page = int(ceil(num_jobs / items_per_page))
+        page_range = range(1, last_page + 1)
+        offset = items_per_page * (page - 1)
+        job_ids = registry.get_job_ids(offset, offset + items_per_page - 1)
+
+        for job_id in job_ids:
+            try:
+                job = Job.fetch(job_id, connection=queue.connection)
```
selwin: We should use `Job.fetch_many`, it's much more efficient for fetching a large number of jobs, and there's no need to handle `NoSuchJobError`

Yolley: Okay, changed fetch in `scheduled_jobs` view, did not touch other views.

django_rq/views.py
```diff
@@ -144,6 +149,43 @@ def failed_jobs(request, queue_index):
...
+    page = int(request.GET.get('page', 1))
+    jobs = []
+
+    if num_jobs > 0:
+        last_page = int(ceil(num_jobs / items_per_page))
+        page_range = range(1, last_page + 1)
+        offset = items_per_page * (page - 1)
+        job_ids = registry.get_job_ids(offset, offset + items_per_page - 1)
+
+        jobs = Job.fetch_many(job_ids, connection=queue.connection)
```
selwin: If one of the jobs is no longer in Redis, `None` will be returned. In this case we should remove the job from registry. Will need a test case for this.

Yolley: I am not so sure, that the view should clear the registry, but ok. And I guess, that this logic should be shared by other views at some point.

PR #404, opened 2020-02-07
Added AdminSite.each_context() to context_data dicts in views

tom-price:
Django's [AdminSite.each_context()](https://docs.djangoproject.com/en/dev/ref/contrib/admin/#django.contrib.admin.AdminSite.each_context) method adds context like the site_header that's used by the admin templates to render the view. It's currently missing from all of the `context_data` dicts provided to the render function. This is why the view still says "Django Administration" even if you've adjusted your site's name. 
missing context
with context

This is also problematic as some 3rd party skins / themes for Django expect to be able to render their templates with information from the request.context, which is missing. I ran into this with the `available_apps` list in Django Suit.

### DIFF
--- a/django_rq/views.py
+++ b/django_rq/views.py
@@ -2,7 +2,7 @@ from __future__ import division
 
 from math import ceil
 
-from django.contrib import messages
+from django.contrib import admin, messages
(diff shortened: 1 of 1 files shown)


selwin:
Thanks!

PR #405, opened 2020-02-12
Allow worker to run with scheduler enabled

stlk:
I added support for ` --with-scheduler` based on https://github.com/rq/rq/blob/master/docs/docs/scheduling.md#running-the-scheduler


selwin:
Thanks!

PR #420, opened 2020-05-08
Fix ignored DEFAULT_RESULT_TTL

fdev:
This library has a very useful setting to set a default `result_tt`, but it is not always respected.

Two issues with the `DEFAULT_RESULT_TTL` setting should be fixed with this PR:
1. `DEFAULT_RESULT_TTL` is ignored in the `job` decorator when it is set to zero.
2. `DEFAULT_RESULT_TTL` is completely ignored in the `enqueue` method.

### DIFF
--- a/django_rq/decorators.py
+++ b/django_rq/decorators.py
@@ -33,7 +33,7 @@ def job(func_or_queue, connection=None, *args, **kwargs):
 
     RQ = getattr(settings, 'RQ', {})
     default_result_ttl = RQ.get('DEFAULT_RESULT_TTL')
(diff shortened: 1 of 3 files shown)


selwin:
Can you add a test for this?

fdev:
I have reverted the fix for the second issue, since it's not a bug but a missing feature. According to the documentation only the `@job` decorator will respect `DEFAULT_RESULT_TTL`, not the `enqueue` method.

A test has been added for the fix of the first issue.

selwin:
Thanks!

PR #428, opened 2020-06-03
Admin improvement: add job.meta to job_detail.html

atten:
This can be useful for debugging/maintenance, why not to include meta field?


selwin:
Thanks!

PR #434, opened 2020-07-06
README: add @job arguments usage

dash-rai:
When specifying arguments to the underlying RQ job decorator, the
queue name must be given explicitly which may not be obvious because
of the simple @job usage.


selwin:
Thanks!

Issue #427, opened 2020-05-27
Set Sentry environment from config variable in job


ttinjic:
Hello,

When I running jobs, Sentry tags : `environment` shows nothing and `server_name` our localhost name even with `environment` and `server_name` arguments in `sentry_sdk.init`.

config :

```
sentry_sdk.init(
 dsn=SENTRY_DSN,
 integrations=[DjangoIntegration(), RqIntegration(), sentry_logging],
 send_default_pii=True,
 max_breadcrumbs=50,
 environment=SENTRY_ENVIRONMENT,
 server_name=SENTRY_ENVIRONMENT,
)
```

when I look at the value of the env var `SENTRY_ENVIRONMENT`, I get
`INFO 2020-05-27 15:00:43,666 [statistics] SENTRY_ENVIRONMENT:tti.test`

however `environment` show nothing and `server_name` tag still show hostname

Do you know how to use the proper environment in a job?

Thanks

selwin:
Not too sure about this, sorry.

PR #442, opened 2020-08-22
Update rqworker command Sentry configuration (alternative version)

### LINKED ISSUE #440: Use existing Sentry client if configured
See #427 for background.

---

If Sentry has been configured in the Django settings (using `sentry_sdk.init`) the Sentry client configuration will be overwritten when running the `rqworker` command.

### Steps to reproduce

1. Set up a Django project that uses django-rq and sentry-sdk
1. Configure sentry in the Django settings file, passing in options (anything other than `debug` or `ca-certs`)
1. Run `./manage.py rqworker`

### Expected outcome

Sentry is configured using the settings applied in the Django settings

### Actual outcome

Sentry is configured with `debug` and `ca-certs` options, but nothing else.

hugorodgerbrown:
This is a simpler alternative to #441. Fixes #440 and #427. 

---

This PR changes how the `rqworker` command integrates with Sentry.

At the current time, if you pass in `sentry-dsn`, **or** have `django.conf.settings.SENTRY_DSN` set, then the command will reinitialise Sentry using a limited subset of options, overriding any existing Django setup. The options supported are `debug` and `ca_certs`, and the list of integrations is restricted to `RqIntegration`. 

This PR prevents the possible clash with Django settings by removing the fallback to `SENTRY_DSN`. If you pass `sentry-dsn` to the command, it will use it, and ignore any existing Django settings configuration. If you don't pass it, it will do nothing - i.e. it will rely on any existing Django configuration.

In addition, this does away with the `rq.contrib.sentry.register_sentry` function call in favour of a direct `sentry_sdk.init` call. The RQ function explicitly sets up Sentry with `integrations=[RqIntegration()]`, preventing the use of any other integrations. This makes sense within the context of `rq`, but if you are using `django-rq` then it makes more sense to use `[RedisIntegration(), RqIntegration(), DjangoIntegration()]`.

This PR will be a breaking change for anyone who does not pass `sentry-dsn` to the command, but **does** have `django.conf.settings.SENTRY_DSN` set. This may be a dealbreaker, but that group seems like an edge case?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ django_rq.egg-info/
 *.rdb
 # pycharm
 .idea
+Pipfile.lock
--- a/Pipfile
+++ b/Pipfile
@@ -8,6 +8,7 @@ django-redis = {editable = true,git = "https://github.com/niwinz/django-redis.gi
 docutils = "*"
 pygments = "*"
 rq-scheduler = "*"
+coverage = "*"
 
 [packages]
 django-rq = {editable = true,path = "."}
--- a/README.rst
+++ b/README.rst
@@ -331,7 +331,7 @@ setting the following Sentry options:
     {
         'debug': options.get('sentry_debug'),
         'ca_certs': options.get('sentry_ca_certs'),
-        'integrations': [RqIntegration(), DjangoIntegration()]
+        'integrations': [RedisIntegration(), RqIntegration(), DjangoIntegration()]
     }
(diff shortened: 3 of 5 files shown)


selwin:
Hey, sorry for the late reply. I’m ok with this change.

selwin:
Give me a few days to further process this to think of other potential ramifications.

PR #458, opened 2020-12-03
Option to update job timeout according to settings

eriktelepovsky:
Ability to change job timeout from its detail. Useful when settings are changed.


selwin:
Sorry for the late reply. It would be much more useful if the user himself can specify the timeout. Can you add an intermediary step where the user can specify the new desired timeout?

rsoltys:
Update of timeout according to the new setting would be appreciated.

PR #468, opened 2021-03-16
Allow setting ssl_cert_reqs and username in get_redis_connection

jeyang:
Hi, I'm a bit new to contributing to a project, happy to make changes as suggested.

Context: This was prompted by not being able to connect to Redis on Heroku.
https://help.heroku.com/HC0F8CUS/redis-connection-issues

I needed to be able to set the ssl_cert_reqs.

Also, question: Should we add all the params below instead of a select few?
https://github.com/andymccurdy/redis-py/blob/1a41cfd95a53a95b078084d8627be6b6fba3bb71/redis/client.py#L737-L749

Thanks!

### DIFF
--- a/django_rq/queues.py
+++ b/django_rq/queues.py
@@ -112,7 +112,7 @@ def get_redis_connection(config, use_strict_redis=False):
             service_name=config['MASTER_NAME'], redis_class=redis_cls,
         )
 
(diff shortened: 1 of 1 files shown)


selwin:
Thanks!

michaelhays:
Hi @selwin, are you able to push out a release with this change?

selwin:
@michaelhays 2.4.1 is now on PyPI :)

PR #469, opened 2021-03-23
sync mode test example to support python 3

antsmc2:
dictionary.itervalues() was removed in python 3. Updated sync mode test example to use dictionary.values() instead.


selwin:
Thanks!

PR #470, opened 2021-03-24
Fix module django_rq has no attribute get

mouadennasri:
A small change here, `get` is not a `django_rq` attribute, it should be `get_queue`.


selwin:
Thanks!

PR #474, opened 2021-04-07
removed link to django-suit-rq

guettli:
At the bottom of the README was a link to: https://github.com/coremke/django-suit-rq

this project was not updated since 2016.

I guess it makes sense to remove this link from the README


selwin:
Thanks!

Issue #479, opened 2021-04-28
django template tag '|escape' fail to escape HTML when string is marked as SafeString


rhenanbartels:
Hi!

First of all, I would like to thank you for the great work with django-rq. It makes the process of creating asynchronous tasks a lot easier.

Currently, I am enqueuing a task using the method `enqueue_in` from **rq/django-rq** and one of the kwargs of the function is a rendered HTML template. The idea is to get some variables, add them to an HTML template and send it via a scheduled e-mail.

Here a [gist](https://gist.github.com/rhenanbartels/b5d2f055ee8bca754a9c974201fbf1ae) with the sending email code.

Because I am using the class `Template` from `django.template` to create the e-mail body, the resulting HTML is a string **already marked as safe** (`django.utils.safestring.SafeString`). For this reason, when checking the results on `django-rq` dashboard, more specifically in the [**job_detail view**](https://github.com/rq/django-rq/blob/83f7e75975c65ae9be553af7fa3483979dfe14d9/django_rq/templates/django_rq/job_detail.html#L150), the html is not escaped - the |escape tag checks if the string has the `__html__` method ([here](https://github.com/django/django/blob/ca9872905559026af82000e46cde6f7dedc897b6/django/template/defaultfilters.py#L413) and [here](https://github.com/django/django/blob/ca9872905559026af82000e46cde6f7dedc897b6/django/utils/html.py#L101)) and returns the raw text. Therefore, any JS script in the HTML is executed when opening the job detail page in admin.

Addiytionally, the `args` of the enqueued function are not escaped:

https://github.com/rq/django-rq/blob/83f7e75975c65ae9be553af7fa3483979dfe14d9/django_rq/templates/django_rq/job_detail.html#L131

This is a [short video](https://youtu.be/0whyOvkSxjc) showing the JS execution in admin.

**VERSIONS**

Python: 3.8.1
Django: 3.1.6
django-rq: 2.3.2

Thanks in advance and congratulations for the awesome project!
[]s

PR #480, opened 2021-04-28
Create force_escape template tag to escape text marked as safe

rhenanbartels:
Hi, this PR tries to solve issue #479 .

A new template tag `force_escape` is created and used in `job_detail.html` to escape `args` and `kwargs` of a job function event if they are `SafeString`


selwin:
Thanks!

PR #482, opened 2021-04-29
Add ability to pass arbitrary Redis client kwargs

juanjgarcia:
Adds the ability to create a dict within the queue definition called `REDIS_CLIENT_KWARGS` of which all kwargs will be passed to the Redis client.

This allows control of timeouts, health check intervals etc. directly with the client.

I broke the line with a small restyle as it was getting very long.


selwin:
Thanks!

PR #506, opened 2021-08-28
Naive implementation of scheduled jobs listing in the admin

mbi:
This is a naive implementation of the scheduled jobs listing  in the admin (#120) using `get_scheduler()`.

It seems to me, all of the ticks mentioned in #121 (most notable efficient pagination) are covered, but alas: as it's a pretty naive implementation and not being familiar with the project, it's certainly missing a few points.

### DIFF
--- a/django_rq/utils.py
+++ b/django_rq/utils.py
@@ -1,16 +1,12 @@
+from django.core.exceptions import ImproperlyConfigured
 from rq.job import Job
-from rq.registry import (
-    DeferredJobRegistry, 
-    FailedJobRegistry, 
-    FinishedJobRegistry, 
-    ScheduledJobRegistry,
-    StartedJobRegistry, 
-    clean_registries
-)
+from rq.registry import (DeferredJobRegistry, FailedJobRegistry,
+                         FinishedJobRegistry, StartedJobRegistry,
+                         clean_registries)
 from rq.worker import Worker
 from rq.worker_registration import clean_worker_registry
 
-from .queues import get_connection, get_queue_by_index
+from .queues import get_connection, get_queue_by_index, get_scheduler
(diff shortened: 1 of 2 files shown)


selwin:
`ScheduledJobRegistry` is actually a newer implementation of RQ's scheduling functionality so we no longer needs this PR. Sorry. I'll close the issue too.

mbi:
Okay, sorry for commenting on a closed issue, but what's the plan to list jobs scheduled with rq-scheduler in django-rq's admin interface?

selwin:
There’s currently no plans to list jobs scheduled with rq-scheduler in
django-rq. If you want to work on it, I suppose you could but it will have
to be done through a different interface.

RQ’s scheduler works on a per queue basis where as there’s only one RQ
scheduler queue/registry for all queues.

mbi:
Okay I think I understand the source of my misunderstanding: the [README](
https://github.com/rq/django-rq/blob/master/README.rst#support-for-rq-scheduler) suggests to schedule tasks using rq-scheduler like this:

```python
import django_rq
scheduler = django_rq.get_scheduler('default')
job = scheduler.enqueue_at(datetime(2020, 10, 10), func)
```
Which results in tasks not appearing in the registry or the admin.

OTOH, this seems an equivalent way of scheduling jobs:

```python
from django_rq.queues import get_queue
queue = get_queue('default')
job = queue.enqueue_at(datetime(2020, 10, 10), func)
```
... which does appear in the registry and the admin.


Am I correct in assuming both methods are equivalent? If yes, shouldn't the README mention the latter instead of the former?

selwin:
@mbi right, that's a good catch. The docs should be updated to use RQ's built in functionality.

PR #512, opened 2021-09-27
Use ModelAdmin class to enable sidebar and permission in django admin

haakenlid:
A different approach to the django admin site integration. Instead of overriding the admin html-template, we can use a unmanaged Queue model and ModelAdmin without a backing database table.

Based on an idea from this Stack Overflow answer
https://stackoverflow.com/a/37988537/1977847

sidebar:
image

regular view on main admin page:
image

Uses Django's contenttypes to enable hiding admin views for users:
image


selwin:
This is a nice PR! Does this also support Django 2.2?

haakenlid:
I don't know if there might be some versions that it doesn't work with. 
I suspect that it would work, since the stackoverflow answer is from 2016.

haakenlid:
> Yeah I wonder why this PR didn't trigger CI. Usually I see a button to start CI for first time contributors but this time I don't see anything. Strange.

There's a message on the travis-ci.org page:  https://travis-ci.org/github/rq/django-rq
> Since June 15th, 2021, the building on travis-ci.org is ceased. Please use travis-ci.com from now on.

Maybe you have to change a config to make it work with the new domain?

Issue #504, opened 2021-08-12
workers count on admin page not reflected correctly


marsha97:
We had 21 workers before, and changed to 16 workers only. We checked the daemon with `systemctl status` and it was 16, but in admin it was  still 21.

selwin:
Yeah, sometimes the worker count may appear out of sync. Worker registries are periodically cleaned up (if I remember correctly once every 15 minutes) by one of RQ's workers so if you're changing the worker count and monitoring it via dashboard, the right worker count may not show up immediately.

The solution is to call [clean_worker_registry(queue)](https://github.com/rq/rq/blob/6bfd47f735de3f297ba3c8f59d5e2dcfa1987107/rq/worker_registration.py#L50) before displaying the worker count in [this view](https://github.com/rq/django-rq/blob/master/django_rq/views.py#L231).

Mine opening a PR for this?

marsha97:
> Yeah, sometimes the worker count may appear out of sync. Worker registries are periodically cleaned up (if I remember correctly once every 15 minutes) by one of RQ's workers so if you're changing the worker count and monitoring it via dashboard, the right worker count may not show up immediately.
> 
> The solution is to call [clean_worker_registry(queue)](https://github.com/rq/rq/blob/6bfd47f735de3f297ba3c8f59d5e2dcfa1987107/rq/worker_registration.py#L50) before displaying the worker count in [this view](https://github.com/rq/django-rq/blob/master/django_rq/views.py#L231).
> 
> Mine opening a PR for this?

Ok noted on this, will try to look into it

selwin:
Fixed in https://github.com/rq/django-rq/pull/516

PR #516, opened 2021-11-07
Sync worker count before displaying on Dashboard

cgl:
Similar to the issue reported in #504 we had some sync issues. Here is the PR of the solution @selwin suggested.

Please do let me know if I can improve in any way.

Also would be nice to get a new release, with all the changes, merged since 2.4.1. @selwin do let me know please if I can be any help.


selwin:
Thanks!

PR #520, opened 2021-11-23
Remove job from ScheduledJobRegistry if enqueued manually

robertaistleitner:
When using admin view to manually enqueue a scheduled job, it is not removed from the ScheduledJobRegistry (I think this was missed when implementing the scheduler).


selwin:
Thanks!

PR #521, opened 2021-12-06
Mentioned RQ built-in scheduler in Readme

martinsvoboda:
Completing Readme - added information about RQ build-in scheduler and highlight that stats display only build-in scheduler jobs.  See #506


selwin:
Thanks!

PR #531, opened 2022-01-10
Remove usage of deprecated LooseVersion

reybog90:
The minimum supported version of Django is now 2.0. All these
checks for 1.9 and 1.10 are no longer needed.

### DIFF
--- a/django_rq/management/commands/rqenqueue.py
+++ b/django_rq/management/commands/rqenqueue.py
@@ -1,7 +1,4 @@
-from distutils.version import LooseVersion
-
 from django.core.management.base import BaseCommand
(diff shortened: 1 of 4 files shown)


awbacker:
Why is this not merged yet?  I'm happy I did a search before making the same PR you already did.  

As to how I got here... On py3.10.4+osx, sometimes, when Django imports `LooseVersion()` in `jquery_path.py` it will hang, permanently.  So far this only happens locally, thank gosh.  I have o idea why it does this, but when I saw it was for such an old unsupported django I figured might as well remove it and save the pain.

selwin:
Thanks @reybog90 !

PR #545, opened 2022-08-23
Add max-jobs option in command

arpit-goel:
Allow `rqworker` management command to accept `max-jobs` option which RQ provide to limit the number of jobs to execute.

### DIFF
--- a/django_rq/management/commands/rqworker.py
+++ b/django_rq/management/commands/rqworker.py
@@ -79,6 +79,8 @@ class Command(BaseCommand):
                             help='A path to an alternative CA bundle file in PEM-format')
         parser.add_argument('--sentry-debug', action='store', default=False, dest='sentry_debug',
                             help='Turns debug mode on or off.')
+        parser.add_argument('--max-jobs', action='store', default=None, dest='max_jobs', type=int,
+                            help='Maximum number of jobs to execute')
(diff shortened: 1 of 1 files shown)


django_rq/management/commands/rqworker.py
```diff
@@ -123,7 +125,10 @@ def handle(self, *args, **options):
             # Close any opened DB connection before any fork
             reset_db_connections()
 
-            w.work(burst=options.get('burst', False), with_scheduler=options.get('with_scheduler', False), logging_level=level)
+            w.work(
+                burst=options.get('burst', False), with_scheduler=options.get('with_scheduler', False),
+                logging_level=level, max_jobs=options['max_jobs']
+            )
```
ushuz, line 131: Probably argument values should be retrieved in a consistent way here? Also, since default values are already defined in argument parser, no need to use `options.get()`.

```suggestion
            w.work(burst=options['burst'], with_scheduler=options['with_scheduler'], max_jobs=options['max_jobs'], logging_level=level)
```

arpit-goel: @ushuz I agree it should be used in a consistent way. Although, `options.get()` is an existing implementation, I have just added `max_jobs` parameter. If you want to follow a particular way, please let me know will change it accordingly.

PR #546, opened 2022-08-31
Support statistics runtime of finished job

yyaadet:
I add a new command named rq_runtime to display runtime of functions. And inset a new column ended in the jobs list page.

### DIFF
--- a/django_rq/management/commands/rq_runtime.py
+++ b/django_rq/management/commands/rq_runtime.py
@@ -1,0 +1,69 @@
+import django_rq
+
+from django.core.management.base import CommandParser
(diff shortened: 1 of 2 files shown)


selwin:
@yyaadet I cherry-picked your commit that shows `job.ended_at` in this https://github.com/rq/django-rq/pull/557 . I'm not sure about iterating through all jobs in `FinishedJobRegistry` and averaging it on the fly.

If we were to have registry level statistics, I think we should build it into RQ proper, like [Worker statistics](https://github.com/rq/rq/blob/master/rq/worker.py#L782).

Issue #567, opened 2023-02-03
Add "Scheduler is active" column for each queue


gabriels1234:
Hi, I'd like to know, now that the user of workers --with-scheduler if it's a good Idea to add a column for "Scheduler active" for each row (queue).
I might be able to help with a PR here.
Thanks!
(cc/ @selwin )

selwin:
Hi, yes PR for this is welcome!

PR #570, opened 2023-02-07
Feature/add scheduler pid info

gabriels1234:
Based off @selwin encouragement (#567), This PR adds to the general view the PID of the scheduler.
It handles both rq-scheduler and rq worker --with-scheduler.

it helps visualize whether there's an issue (e.g worker started without scheduler, or rq-scheduler not responding). Since worker-schedulers are queue-specific, it made a lot of sense to add it as a column to the main django-rq view.

It does that by checking scheduler data in redis (getting connection either from the queue or from the scheduler) and considers that rq and rq-scheduler store this differently.

If there's no active scheduler, the column will show "Inactive" (this might take some time since the data in redis has to expire (1~(rq) to 15~ (rq-scheduler) minutes)).

Wrote some tests, but need help in running them (rq has it easier). 
PS: (PIDs can be the same but the schedulers can be different (check the name of the worker to confirm they are different))

### DIFF
--- a/django_rq/models.py
+++ b/django_rq/models.py
@@ -13,6 +13,7 @@ if not get_commit_mode():
 class Queue(models.Model):
     """Placeholder model with no database table, but with django admin page
     and contenttype permission"""
+    id = models.AutoField(primary_key=True)
 
     class Meta:
         managed = False  # not in Django's database
--- a/django_rq/models.py
+++ b/django_rq/models.py
@@ -13,8 +13,6 @@ if not get_commit_mode():
 class Queue(models.Model):
     """Placeholder model with no database table, but with django admin page
     and contenttype permission"""
-    id = models.AutoField(primary_key=True)
-
     class Meta:
         managed = False  # not in Django's database
         default_permissions = ()
--- a/django_rq/tests/tests.py
+++ b/django_rq/tests/tests.py
@@ -808,8 +808,6 @@ class TemplateTagTest(TestCase):
         self.assertEqual(escaped_string, expected)
 
 
-def mocked_relase_locks():
-    print("Mock Release Locks")
 class SchedulerPIDTest(TestCase):
     @skipIf(RQ_SCHEDULER_INSTALLED is False, 'RQ Scheduler not installed')
     def test_scheduler_scheduler_pid_active(self):
--- a/django_rq/tests/tests.py
+++ b/django_rq/tests/tests.py
@@ -811,7 +811,6 @@ class TemplateTagTest(TestCase):
 def mocked_relase_locks():
     print("Mock Release Locks")
 class SchedulerPIDTest(TestCase):
-
     @skipIf(RQ_SCHEDULER_INSTALLED is False, 'RQ Scheduler not installed')
     def test_scheduler_scheduler_pid_active(self):
         test_queue = 'scheduler_scheduler_active_test'
@@ -826,7 +825,9 @@ class SchedulerPIDTest(TestCase):
         with patch('django_rq.utils.QUEUES_LIST',
             new_callable=PropertyMock(return_value=queues)):
             scheduler = get_scheduler(test_queue)
-            self.assertIsNotNone(get_scheduler_pid(get_queue(scheduler.queue_name)))  # No queue object needed
+            scheduler.register_birth()
+            self.assertIsNotNone(get_scheduler_pid(get_queue(scheduler.queue_name)))
+            scheduler.register_death()
     
     @skipIf(RQ_SCHEDULER_INSTALLED is False, 'RQ Scheduler not installed')
     def test_scheduler_scheduler_pid_inactive(self):
@@ -841,10 +842,12 @@ class SchedulerPIDTest(TestCase):
         }]        
         with patch('django_rq.utils.QUEUES_LIST',
             new_callable=PropertyMock(return_value=queues)):
+            connection = get_connection(test_queue)
+            connection.flushall()  # flush is needed to isolate from other tests
             scheduler = get_scheduler(test_queue)
-            scheduler.register_death()  # will mark the scheduler as death so get_scheduler_pid will return None
             scheduler.remove_lock()
-            self.assertIsNone(get_scheduler_pid(queue=None))  # No queue object needed
+            scheduler.register_death()  # will mark the scheduler as death so get_scheduler_pid will return None
+            self.assertIsNone(get_scheduler_pid(get_queue(scheduler.queue_name)))
 
     @skipIf(RQ_SCHEDULER_INSTALLED is True, 'RQ Scheduler installed (no worker--with-scheduler)')
     def test_worker_scheduler_pid_active(self):
--- a/django_rq/tests/tests.py
+++ b/django_rq/tests/tests.py
@@ -826,7 +826,7 @@ class SchedulerPIDTest(TestCase):
             new_callable=PropertyMock(return_value=queues)):
             scheduler = get_scheduler(test_queue)
             scheduler.register_birth()
-            self.assertIsNotNone(get_scheduler_pid(get_queue(scheduler.queue_name)))
+            self.assertIs(get_scheduler_pid(get_queue(scheduler.queue_name)), False)
             scheduler.register_death()
     
     @skipIf(RQ_SCHEDULER_INSTALLED is False, 'RQ Scheduler not installed')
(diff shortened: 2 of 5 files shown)


django_rq/utils.py
```diff
@@ -10,9 +10,41 @@
...
+        Note: result might have some delay (1-15 minutes) but it helps visualizing whether the setup is working correcly
+    '''
+    try:
+        # first try to use rq-scheduler
+        scheduler = get_scheduler()  # should fail if rq_scheduler not present
+        # currently, scheduler_lock_key is not used but, just in case, try
+        lock_key = scheduler.scheduler_lock_key
+        with scheduler.connection.pipeline() as p:
+            if _ := p.get(lock_key):
+                return scheduler.pid
```
gabriels1234, line 32: For some reasons the lock_key was not always there.
So, the extra check for the scheduler instance.

django_rq/utils.py
```diff
@@ -10,9 +10,41 @@
...
+            else:
+                for key in p.keys(f"{scheduler.redis_scheduler_namespace_prefix}*"):
+                    if not p.hexists(key, 'death'):
+                        return scheduler.pid
```
gabriels1234, line 36: For some reasons the lock_key was not always there.
So, the extra check for the scheduler instance.

selwin:
@gabriels1234 do you mind checking the tests and make sure they pass?

gabriels1234:
> @gabriels1234 do you mind checking the tests and make sure they pass?

@selwin Thanks, made it so the tests work. 2 tests with --with-scheduler and 2 tests with RQ-Scheduler. (Tests passed via installing rq-scheduler, testing (2 passed, 2 skipped), then uninstalling rq-scheduler, testing (2 passed, 2 skipped))

gabriels1234:
@selwin let me know if the tests work for you. (I added 4 new tests, 2 (active/inactive) for --with-scheduler and 2 for rq-scheduler). Not sure if the other tests were working or not, but mine do work.

gabriels1234:
@selwin Based on the merged PR by dear @cunla (https://github.com/rq/rq/pull/1787), now that RQ proper has the scheduler_pid, should I use that and add it to the logic to get the pid when the scheduler is rq-scheduler?

cunla:
> @selwin Based on the merged PR by dear @cunla ([rq/rq#1787](https://github.com/rq/rq/pull/1787)), now that RQ proper has the scheduler_pid, should I use that and add it to the logic to get the pid when the scheduler is rq-scheduler?

It needs to be released prior to that. @selwin, when is it expected to be released?

selwin:
> @selwin Based on the merged PR by dear @cunla (https://github.com/rq/rq/pull/1787), now that RQ proper has the scheduler_pid, should I use that and add it to the logic to get the pid when the scheduler is rq-scheduler?

Let's just add a note that the current implementation should eventually be replaced with `queue.scheduler_pid` a few months after RQ 1.13 is released. Currently this library only [requires RQ >= 1.2](https://github.com/rq/django-rq/blob/master/setup.py#L17) so bumping it all the way up to require 1.13 is a bit too big of a jump. I'll move the minimum RQ requirement to 1.8 in the next release so we can slowly drop support for older RQ versions.

django_rq/utils.py
```diff
@@ -10,9 +10,42 @@
...
+
+        # first try to use rq-scheduler
+        scheduler = get_scheduler(queue.name)  # should fail if rq_scheduler not present
+        # currently, scheduler_lock_key is not used but, just in case, try
+        lock_key = scheduler.scheduler_lock_key
+        conn = scheduler.connection
+        if _ := conn.get(lock_key):
+            return scheduler.pid
+        else:
+            for key in conn.keys(f"{scheduler.redis_scheduler_namespace_prefix}*"):
```
selwin, line 37: I just realized that running `keys()` will be very costly in a Redis database with a large number of keys. I think we should avoid doing this and just support the built in `RQScheduler`.

django_rq/utils.py
```diff
@@ -10,9 +10,42 @@
...
+
+def get_scheduler_pid(queue):
+    '''Checks whether there's a scheduler-lock on a particular queue, and returns the PID.
+        It works by first checking if there's an RQ-Scheduler active.
+        If not, it checks the RQ's RQScheduler for a scheduler lock in the desired queue
+        Note: result might have some delay (1-15 minutes) but it helps visualizing whether the setup is working correcly
+    '''
+    try:
+        if not queue:
+            raise ValueError("queue argument not defined")
```
selwin, line 27: We can remove this check since `queue` is already a required argument.

django_rq/models.py
```diff
@@ -13,6 +13,7 @@
 class Queue(models.Model):
     """Placeholder model with no database table, but with django admin page
     and contenttype permission"""
+    id = models.AutoField(primary_key=True)
```
selwin, line 16: What's this used for?

django_rq/utils.py
gabriels1234: Yes, I agree totally. Perhaps making a PR to make sure rq-scheduler sets scheduler_lock_key? because there will be a transition period. for now, I will make it so if no scheduler_lock_key, will return -1 and will be rendered as "Unknown". Will let you know after I finish fixing the PR.

django_rq/models.py
gabriels1234: I removed it. it is causing a Warning, but since it's nor related to this PR I removed it. Thanks.

django_rq/utils.py
gabriels1234: Great! Done.

django_rq/utils.py
gabriels1234: Finally, I decided to just omit functionality if RQ-Scheduler is being used as per your recommendation.

django_rq/tests/tests.py
```diff
@@ -808,6 +808,88 @@ def test_force_escape_regular_string(self):
         self.assertEqual(escaped_string, expected)
 
 
+def mocked_relase_locks():
```
selwin, line 811: This should be deleted.

selwin:
May I get a screenshot of how this looks like on a queue with scheduler active?

django_rq/utils.py
```diff
@@ -10,9 +10,32 @@
...
+    '''Checks whether there's a scheduler-lock on a particular queue, and returns the PID.
+        It Only works with RQ's Built-in RQScheduler.
+        When RQ-Scheduler is available returns False 
+        If not, it checks the RQ's RQScheduler for a scheduler lock in the desired queue
+        Note: result might have some delay (1-15 minutes) but it helps visualizing whether the setup is working correcly
+    '''
+    try:
+        # first try get the rq-scheduler
+        scheduler = get_scheduler(queue.name)  # should fail if rq_scheduler not present
+        return False  # Not possible to give useful information without creating a performance issue (redis.keys())
```
selwin, line 29: This will cause `get_scheduler_pid` to always return `False` if `rq_scheduler` is installed, even though the built in scheduler is used.

I think we should just skip the check for external scheduler altogether.

gabriels1234: Well, ok. In my tests, running both --with-scheduler and rq-scheduler doesn't end up well.
By returning False, I'm basically _disabling_ the feature, since it won't be shown in the dashboard.

After Some digging I found that, for some reason, rq and rq-scheduler organize scheduling and enqueuing of scheduled jobs in parallel ways, namely incompatible.

Please correct me if I'm wrong: Basically, jobs scheduled using Scheduler.enqueue_at will never be picked up by a worker --with-scheduler (*different* redis keys), and also the opposite is true. I wouldn't understand anyone having both running (and if they do, the worker will be worker only).

Recently, @cunla's django-rq-scheduler (>2022.12.1) transitioned to --with-scheduler.

My point is: showing that the Scheduler is active (worker --with-scheduler) and then scheduling jobs using rq-scheduler enqueue_at will cause stale jobs that never run and are never seen by django-rq.

What I propose is your original idea but only working if the rq-scheduler is not to be found, otherwise, the information would be misleading, to say the least.

Let me know

django_rq/tests/tests.py
gabriels1234: Fixed!

django_rq/utils.py
gabriels1234: See below, functionality was removed for rq-scheduler.

django_rq/utils.py
gabriels1234: @selwin As per my 

> What I propose is your original idea but only working if the rq-scheduler is not to be found, otherwise, the information would be misleading, to say the least.

@selwin as I mentioned, I prefer not to activate the feature for rq-scheduler, because it would be misleading.
I think the PR is ready to move forward. (At least to ship it as a half-feature)

However, I'm happy to make it according to your opinion.

gabriels1234:
@selwin looking forward to hear from you how should I go about.

django_rq/utils.py
selwin: I agree, let's drop `rq-scheduler` altogether as it will only make things confusing.

gabriels1234: @selwin great! So the PR is good as it is. do you need to see a screenshot?

selwin: Yes please, a screenshot would be good just to make sure that it works :)

selwin:
Sorry for the late reply @gabriels1234

gabriels1234:
Thanks!
image
Last column

Issue #573, opened 2023-02-27
How to stop/terminate a long-running job. Deleting the job doesn't do the trick


gabriels1234:
Hi,

I was wondering if django-rq provides a way to "terminate" a running job.
For Jobs that take hours, deleting the job would not make the trick. or is it that it's a limitation from RQ?

Thanks!

gabriels1234:
Any guesses?
@selwin Should I investigate and try to make it happen if possible? (i.e adding an option to "stop/terminate" a job?)

selwin:
@gabriels1234 sorry I just saw this. RQ actually provides a way to stop a job that's currently executing. https://python-rq.org/docs/jobs/#stopping-a-currently-executing-job

My advice would be to add a button to send a stop job command in job/worker detail page.

PR #577, opened 2023-03-19
Add Stop Jobs action

gabriels1234:
Adds #573, action to stop a job. Current behavior deletes the entry but does not stop the job. 
@selwin

### DIFF
--- a/django_rq/templates/django_rq/job_detail.html
+++ b/django_rq/templates/django_rq/job_detail.html
@@ -187,7 +187,7 @@
     <div class="submit-row">
         <p class="deletelink-box"><a href="delete/" class="deletelink">Delete</a></p>
         {% if job.is_started %}
-            <form method = 'POST' action = "{% url 'rq_requeue_job' queue_index job.id %}">
+            <form method = 'POST' action = "{% url 'rq_stop_job' queue_index job.id %}">
                 {% csrf_token %}
                 <input type="submit" value="Stop" class="deletelink" name="stop">
             </form>
--- a/django_rq/templates/django_rq/job_detail.html
+++ b/django_rq/templates/django_rq/job_detail.html
@@ -186,6 +186,12 @@
 
     <div class="submit-row">
         <p class="deletelink-box"><a href="delete/" class="deletelink">Delete</a></p>
+        {% if job.is_started %}
+            <form method = 'POST' action = "{% url 'rq_requeue_job' queue_index job.id %}">
+                {% csrf_token %}
+                <input type="submit" value="Stop" class="deletelink" name="stop">
+            </form>
+        {% endif %}
         {% if job.is_failed %}
             <form method = 'POST' action = "{% url 'rq_requeue_job' queue_index job.id %}">
                 {% csrf_token %}
--- a/django_rq/urls.py
+++ b/django_rq/urls.py
@@ -30,4 +30,7 @@ urlpatterns = [
     re_path(
         r'^queues/(?P<queue_index>[\d]+)/(?P<job_id>[-\w\.\:\$]+)/enqueue/$', views.enqueue_job, name='rq_enqueue_job'
     ),
+    re_path(
+        r'^queues/(?P<queue_index>[\d]+)/(?P<job_id>[-\w\.\:\$]+)/stop/$', views.stop_job, name='rq_stop_job'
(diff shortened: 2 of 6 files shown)


gabriels1234:
@selwin Please let me know if it needs any change.
Thanks!

selwin:
Hey, I was thinking we should add the `Stop` button on job detail page like this, along with other buttons (screenshot attached).

CleanShot 2023-03-23 at 15 59 18

django_rq/views.py
```diff
@@ -493,6 +494,17 @@ def actions(request, queue_index):
                 for job_id in job_ids:
                     requeue_job(job_id, connection=queue.connection)
                 messages.info(request, 'You have successfully requeued %d  jobs!' % len(job_ids))
+            elif request.POST['action'] == 'stop':
+                cancelled_jobs = 0
+                for job_id in job_ids:
+                    try:
+                        send_stop_job_command(queue.connection, job_id)
+                        job = Job.fetch(job_id, connection=queue.connection)
```
selwin, line 502: You shouldn't need to manually cancel the job. Does RQ not already automatically put the job in `FailedJobRegistry`?

If that's not the case this should be a bug that needs to be fixed within RQ.

gabriels1234: @selwin Well, I was struggling a bit with the tests (to make it actually catch the job before it's marked as finished). 
But also, if you stop the job, you might not want to have the job in the failed column. I will be doing some tests so I'm happy to hold a bit. Also I will add the stop button that you suggested.

django_rq/templates/django_rq/job_detail.html
```diff
@@ -186,6 +186,12 @@
 
     <div class="submit-row">
         <p class="deletelink-box"><a href="delete/" class="deletelink">Delete</a></p>
+        {% if job.is_started %}
+            <form method = 'POST' action = "{% url 'rq_requeue_job' queue_index job.id %}">
```
selwin, line 190: This does not look like it points to the correct URL.

selwin: @gabriels1234 let me know if you can fix this, I'd like to merge this in time for the next release :)

django_rq/urls.py
```diff
@@ -30,4 +30,7 @@
     re_path(
         r'^queues/(?P<queue_index>[\d]+)/(?P<job_id>[-\w\.\:\$]+)/enqueue/$', views.enqueue_job, name='rq_enqueue_job'
     ),
+    re_path(
+        r'^queues/(?P<queue_index>[\d]+)/(?P<job_id>[-\w\.\:\$]+)/stop/$', views.stop_job, name='rq_stop_job'
```
selwin, line 34: Mind updating the regex for `job_id` to match the newly updated one in this PR? https://github.com/rq/django-rq/pull/586/files

django_rq/views.py
```diff
@@ -529,3 +536,20 @@ def enqueue_job(request, queue_index, job_id):
         'queue': queue,
     }
     return render(request, 'django_rq/delete_job.html', context_data)
+
+
+@never_cache
+@staff_member_required
+def stop_job(request, queue_index, job_id):
+    """Stop started job"""
+    if request.method == 'POST':
```
selwin, line 545: We can use `require_POST` view decorator so we don't need to check for the method. https://docs.djangoproject.com/en/4.2/topics/http/decorators/#django.views.decorators.http.require_POST

django_rq/templates/django_rq/job_detail.html
gabriels1234: Done.

django_rq/views.py
gabriels1234: @selwin I fixed as per your requests, stop job does not cancel it. it just goes to the Failed registry.
Also the button was added.
I think it's ready to be merged.

django_rq/urls.py
gabriels1234: Done.

django_rq/views.py
gabriels1234: Done.

gabriels1234:
@selwin let me know if more changes are needed for this to be merged.
Thanks!

selwin:
Thanks!

PR #581, opened 2023-04-02
Add support for rqworker's custom Serializer

gabriels1234:
@selwin adding support for custom serializers (trying to keep up with rq worker arguments that are not available for the rqworker command). Tests included.


selwin:
Thanks!

PR #583, opened 2023-04-04
User connection context manager for pushing connection to local stack

Cerebro92:


django_rq/management/commands/rqworker.py
```diff
@@ -119,16 +119,16 @@ def handle(self, *args, **options):
...
             # without this, jobs using RQ's get_current_job() will fail
-            use_connection(w.connection)
-            # Close any opened DB connection before any fork
-            reset_db_connections()
-
-            w.work(
-                burst=options.get('burst', False), with_scheduler=options.get('with_scheduler', False),
-                logging_level=level, max_jobs=options['max_jobs']
-            )
+            with Connection(w.connection):
```
selwin, line 124: Can we get rid of this altogether?

PR #586, opened 2023-04-24
More permissive urls for customized job_ids

dmclain:
Virtually any string is a valid rq job_id. I think the only real requirement is that it be a valid redis key, and Redis supports arbitrary binary keys. This PR expands the url patterns so that fewer customized job_ids cause a 500 while viewing the queue.


selwin:
Thanks!

PR #590, opened 2023-05-01
Fix Exception Rendering on Job Details Template

BobReid:
## Issue
The stack trace that caused a job to fail was not displayed for failed jobs on the job details admin screen. 
The html block was wrapped in an if statement checking a non-existent variable exc_info instead of job.exc_info

## Fix
Update the if check to check job.exc_info instead of exc_info


BobReid:
I am closing this out as it is caused by an incompatibility with django-rq and and old version of rq (1.7). Upgrading RQ is the real fix.

PR #598, opened 2023-05-04
Change h1 text in stats template

### LINKED ISSUE #580: Page <h1> redundancy in text
While the `browse_queue/` page metatitle is "Queues", the `<h1>` for the same page is "RQ Queues", which is a repetition too ("Q" already stands for "Queue(s)". I feel "Queues" should be the text for the `<h1>` as well - thus without "RQ".

"Redis Queues" could also be appropriate for both.

What do you think?

niccolomineo:
Fixes #580


selwin:
Thanks!

PR #607, opened 2023-06-01
Add template tag items to always call dictionary function

### LINKED ISSUE #606: Admin job detail page error
`job detail` page sometimes return an error when being opened

error:
```
ValueError
Need 2 values to unpack in for loop; got 1. 
```

This probably is the one causing the error, will investigate more on why the error occured
https://github.com/rq/django-rq/blob/a800c2448294981c3c10187a9faad6a87f55793e/django_rq/templates/django_rq/job_detail.html#L141-L145

chromium7:
Fix #606 

By using a custom template tag, we can ensure that the function `items` is called instead of the key inside the dictionary

### DIFF
--- a/django_rq/templates/django_rq/job_detail.html
+++ b/django_rq/templates/django_rq/job_detail.html
@@ -139,7 +139,7 @@
                     {% if data_is_valid %}
                         {% if job.kwargs %}
                             <ul>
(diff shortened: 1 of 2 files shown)


selwin:
Thanks!

PR #609, opened 2023-06-09
Add rqcleanqueue to management commands

rfreis:
Hi guys,
I hope this PR finds you all well!

During my daily usage on local environment faced situations that this command could help a lot - actually I'm already using it on my local environment - and thought could be a good idea share this with everybody else

If you find useful, I'll be happy to contribute with the required changes/upgrades to meet the project standards or further improvements

### DIFF
--- a/django_rq/management/commands/rqcleanqueue.py
+++ b/django_rq/management/commands/rqcleanqueue.py
@@ -1,0 +1,24 @@
+from django.core.management.base import BaseCommand
+
+from ... import get_queue
+
+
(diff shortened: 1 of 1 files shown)


selwin:
Hi @rfreis mind writing a unit test for this? I think it'd also be better to rename this to `rq-empty-queue`. RQ has `clean_registries` method, which does some house keeping tasks, so I don't want to use the `clean` terminology for something that will delete the entire queue.

PR #620, opened 2023-08-23
Update templates to match with current Django Admin

nikhilweee:
This PR updates templates to better match with those of Django 4.2

### DIFF
--- a/django_rq/templates/django_rq/jobs.html
+++ b/django_rq/templates/django_rq/jobs.html
@@ -28,7 +28,7 @@
     <div class="breadcrumbs">
         <a href="{% url 'admin:index' %}">Home</a> &rsaquo;
         <a href="{% url 'rq_home' %}">Django RQ</a> &rsaquo;
-        <a href = "{% url 'rq_jobs' queue_index %}">{{ queue.name }}</a>
+        <a href="{% url 'rq_jobs' queue_index %}">{{ queue.name }}</a>
     </div>
 {% endblock %}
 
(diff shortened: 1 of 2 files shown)


nikhilweee:
@selwin Would be nice to hear your thoughts and hopefully get this merged soon.

selwin:
Hi @nikhilweee mind attaching a screenshot of before and after?

selwin:
And how does this change affect Django < 4.2?

nikhilweee:
This is what `/django_rq/queue/` looked like before

image

This is what `/django_rq/queue/` looks like after this PR

image

---
The difference is subtle here, but is more pronounced in the next template

---

This is what `/django-rq/queues/0/scheduled/` looked like before

image

This is what `/django-rq/queues/0/scheduled/` looks like after this PR

image

nikhilweee:
@selwin My changes are based off of the current [`change_list.html`](https://github.com/django/django/blob/main/django/contrib/admin/templates/admin/change_list.html) template.

I am not sure about the impact on Django < 4.2, but comparing v3.2.20 to v4.2.4 on GitHub, looks like that page hasn't changed. So I am tempted to say it's safe to assume that this PR should be applicable to both Django 3.2 and 4.2?

https://github.com/django/django/compare/4.2.4...3.2.20

selwin:
Do you mind testing this change on Django 3.2 at the very least? We don't have unit tests for this so I'd feel much more comfortable if I can see a screenshot of how it looks on older versions.

nikhilweee:
@selwin Here are results on Django 3.2

`django_rq/queue` Before this PR

image

`django_rq/queue` After this PR

image

---

`django_rq/queues/0/scheduled` Before this PR

image

`django_rq/queues/0/scheduled` After this PR

image

nikhilweee:
Hey @selwin, any thoughts?

selwin:
Thanks!

PR #629, opened 2023-11-16
Delete all failed jobs

chromium7:
Deleting failed jobs in the admin page currently can only be done per page, which can get quite tedious when there are a lot of failed jobs.

https://github.com/rq/django-rq/assets/61144981/9b41a29a-687a-4d84-b76d-2ebbb17e658f


django_rq/views.py
```diff
@@ -451,6 +451,38 @@ def requeue_all(request, queue_index):
...
+    queue = get_queue_by_index(queue_index)
+    registry = FailedJobRegistry(queue=queue)
+
+    if request.method == 'POST':
+        job_ids = registry.get_job_ids()
+        count = 0
+        # Confirmation received
+        for job_id in job_ids:
+            try:
+                job = Job.fetch(job_id, connection=queue.connection)
```
selwin, line 467: Mind using `Job.fetch_many()`?

Issue #626, opened 2023-09-20
Allow Specifying Default Serializer for Django-RQ Queue


sophcass:
## Description
We are currently using `django-rq` alongside Redis Queue (RQ). In our setup, we've opted to use the `JSONSerializer` instead of the default pickle serializer. The reason behind this choice is the potential versioning issues that can arise from using pickle.

## Details
We've successfully configured our setup to use the `JSONSerializer` for processing jobs in the queue. However we've run into an issue when it comes to utilising the `django-rq` admin dashboard while also using the `JSONSerializer`.

The challenge we face is that `django-rq` isn't aware of the serializer we've chosen, and as a result, it doesn't specify the serializer when fetching information to be displayed in the admin dashboard. As a result, the dashboard becomes somewhat limited and less useful for us.

The admin dashboard feature was one of the primary reasons we have adopted `django-rq`, and we are eager to continue leveraging it while also using the `JSONSerializer` for our jobs.

## Proposed Solution
We would like to propose opening a PR to enhance `django-rq`'s functionality. Specifically, to enable users to specify their desired default serializer when they initially set up the Redis queue. This enhancement would simplify our configuration process, ensuring that we don't need to set the serializer in multiple places when setting up the Redis queue via `django-rq` and also it would allow us to utilise the admin dashboard in conjunction with `JSONSerializer`.

If you are happy for us to make these changes, we can get to work on a PR to implement them! Let us know!

sophcass:
@selwin it would be great to hear your thoughts about this proposition, so that we can start working on it.

selwin:
Fixed in https://github.com/rq/django-rq/pull/630

PR #630, opened 2023-11-16
Specify queue serializer when creating Django RQ queue

sophcass:
Related to this issue: https://github.com/rq/django-rq/issues/626 

We want to be able to add the serializer when creating the Django RQ queue. This serializer should then automatically be used when fetching the jobs so that we do not get `DeserializationError` in the Django RQ dashboard.

### DIFF
--- a/django_rq/queues.py
+++ b/django_rq/queues.py
@@ -155,6 +155,7 @@ def get_queue(
     connection=None,
     queue_class=None,
     job_class=None,
+    serializer=None,
     **kwargs
 ):
     """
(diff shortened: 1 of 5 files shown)


sophcass:
Without setting the correct serializer when fetching the job, an Unpickling/Deserialization Error appears in place of the information. See here:
image

This is what the job details page looks like now that the correct serializer is being used:
image

selwin:
Thanks!

sophcass:
Can we get create a release for the couple of changes that have been merged recently?

selwin:
Yes, will try to cut a release sometime in the next few days.

PR #637, opened 2023-12-18
fix: use default redis port when not specified in connection kwargs

terencehonles:
If a port is not specified explicitly `Redis` uses its default port instead, and it may not always appear in the connection kwargs.

Related to https://github.com/rq/django-rq/pull/633


selwin:
Thanks!

Issue #650, opened 2024-02-27
Database Errors When Running rqworker-pool


GhalebKhaled:
I'm trying to switch from multiple workers for different queues into worker-pool, using the command below
```
./manage.py rqworker-pool main default secondary --num-workers 2 --worker-class rq.worker.HerokuWorker --settings settings.rq_settings
```
I'm getting database errors:

```
 File "/app/.heroku/python/lib/python3.11/site-packages/django/db/models/sql/compiler.py", line 1562, in execute_sql
    cursor.execute(sql, params)
  File "/app/.heroku/python/lib/python3.11/site-packages/sentry_sdk/integrations/django/__init__.py", line 641, in execute
    return real_execute(self, sql, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/.heroku/python/lib/python3.11/site-packages/django/db/backends/utils.py", line 67, in execute
    return self._execute_with_wrappers(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/.heroku/python/lib/python3.11/site-packages/django/db/backends/utils.py", line 80, in _execute_with_wrappers
    return executor(sql, params, many, context)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/.heroku/python/lib/python3.11/site-packages/django/db/backends/utils.py", line 84, in _execute
    with self.db.wrap_database_errors:
  File "/app/.heroku/python/lib/python3.11/site-packages/django/db/utils.py", line 91, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "/app/.heroku/python/lib/python3.11/site-packages/django/db/backends/utils.py", line 89, in _execute
    return self.cursor.execute(sql, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
django.db.utils.OperationalError: SSL SYSCALL error: EOF detected
```

Other database error appear if I used the `SimpleWorker`
```
File "/app/.heroku/python/lib/python3.11/site-packages/django/db/models/query.py", line 380, in __len__
    self._fetch_all()
  File "/app/.heroku/python/lib/python3.11/site-packages/django/db/models/query.py", line 1881, in _fetch_all
    self._result_cache = list(self._iterable_class(self))
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #655, opened 2024-03-25
Worker pool handles spawned process

### LINKED ISSUE #651: RQ WorkerPool is not loading models at all
Hi,

This issue is close to #650, but we are not in the same context.

I need my jobs to read/write models on database, so I've made a simple task to check that it's working.

```
@job("data-tasks", timeout=3600)
def Test(instance):
    print(instance)
```

When using a single worker, no issue, I can pass the id, or the object, without any issue.

`python manage.py rqworker data-tasks notifications` 

```
16:49:47 data-tasks: data.methods.Test(<Instance: [MyObject] LVX Vlxx>) (9ea7c2a5-420e-4849-873a-fe378708f760)
[MyObject] LVX Vlxx
16:49:50 data-tasks: Job OK (9ea7c2a5-420e-4849-873a-fe378708f760)
```

But when I'm using a pool of 1 worker, then it seems that something is going wrong, the content of Django App is not loaded.

`python manage.py rqworker-pool data-tasks notifications --num-workers 1` 

```
16:51:36 data-tasks: data.methods.Test(<Instance: [MyObject] LVX Vlxx>) (76b6fb38-8f10-4e78-9231-9fafc7c29042)
16:51:37 [Job 76b6fb38-8f10-4e78-9231-9fafc7c29042]: exception raised while executing (<DeserializationError>)
  File "/Users/mymac/PycharmProjects/MyApp/.venv/lib/python3.10/site-packages/rq/job.py", line 486, in _deserialize_data
    self._func_name, self._instance, self._args, self._kwargs = self.serializer.loads(self.data)
  File "/Users/mymac/PycharmProjects/MyApp/.venv/lib/python3.10/site-packages/django/db/models/base.py", line 2612, in model_unpickle
    model = apps.get_model(*model_id)
  File "/Users/mymac/PycharmProjects/MyApp/.venv/lib/python3.10/site-packages/django/apps/registry.py", line 201, in get_model
    self.check_models_ready()
  File "/Users/mymac/PycharmProjects/MyApp/.venv/lib/python3.10/site-packages/django/apps/registry.py", line 143, in check_models_ready
    raise AppRegistryNotReady("Models aren't loaded yet.")
django.core.exceptions.AppRegistryNotReady: Models aren't loaded yet.

The above exception was the direct cause of the following exception:

  File "/Users/mymac/PycharmProjects/MyApp/.venv/lib/python3.10/site-packages/rq/worker.py", line 1422, in perform_job
    self.prepare_job_execution(job, remove_from_intermediate_queue)
  File "/Users/mymac/PycharmProjects/MyApp/.venv/lib/python3.10/site-packages/rq/worker.py", line 1351, in prepare_job_execution
    self.procline(msg.format(job.func_name, job.origin, time.time()))
  File "/Users/mymac/PycharmProjects/MyApp/.venv/lib/python3.10/site-packages/rq/job.py", line 520, in func_name
    self._deserialize_data()
  File "/Users/mymac/PycharmProjects/MyApp/.venv/lib/python3.10/site-packages/rq/job.py", line 488, in _deserialize_data
    raise DeserializationError() from e
rq.exceptions.DeserializationError
```

I've also tried to pass only the ID and get the object, then I've an error that is telling that `AbstractUser`cannot be loaded.

```
 File "/Users/mymac/PycharmProjects/MyApp/data/models.py", line 4, in <module>
    from django.contrib.auth.models import User
  File "/Users/mymac/PycharmProjects/MyApp/.venv/lib/python3.10/site-packages/django/contrib/auth/models.py", line 3, in <module>
    from django.contrib.auth.base_user import AbstractBaseUser, BaseUserManager
  File "/Users/mymac/PycharmProjects/MyApp/.venv/lib/python3.10/site-packages/django/contrib/auth/base_user.py", line 58, in <module>
    class AbstractBaseUser(models.Model):
  File "/Users/mymac/PycharmProjects/MyApp/.venv/lib/python3.10/site-packages/django/db/models/base.py", line 129, in __new__
    app_config = apps.get_containing_app_config(module)
  File "/Users/mymac/PycharmProjects/MyApp/.venv/lib/python3.10/site-packages/django/apps/registry.py", line 260, in get_containing_app_config
    self.check_apps_ready()
  File "/Users/mymac/PycharmProjects/MyApp/.venv/lib/python3.10/site-packages/django/apps/registry.py", line 138, in check_apps_ready
    raise AppRegistryNotReady("Apps aren't loaded yet.")
django.core.exceptions.AppRegistryNotReady: Apps aren't loaded yet.
``` 

So it seems to me that worker pool is missing something.

Thanks for your help,

chromium7:
Fix #651
Ref #650

I can't reproduce the issue in #650, but I did get the same errors in #651
Seems like its an issue with `multiprocessing` start method in certain OS, as was answered in this stack overflow post: https://stackoverflow.com/questions/46908035/apps-arent-loaded-yet-exception-occurs-when-using-multi-processing-in-django

This solution builds on top of @jackkinsella answer, and can be updated further to handle his specific issue.

### DIFF
--- a/django_rq/worker_pool.py
+++ b/django_rq/worker_pool.py
@@ -1,0 +1,38 @@
+import django
+from multiprocessing import Process, get_start_method
+from typing import Any
+
+from rq.worker_pool import WorkerPool, run_worker
+
+
+class DjangoWorkerPool(WorkerPool):
+    def get_worker_process(
+        self,
+        name: str,
+        burst: bool,
+        _sleep: float = 0,
+        logging_level: str = "INFO",
+    ) -> Process:
+        """Returns the worker process"""
+        return Process(
+            target=run_django_worker,
+            args=(name, self._queue_names, self._connection_class, self._pool_class, self._pool_kwargs),
+            kwargs={
+                '_sleep': _sleep,
+                'burst': burst,
+                'logging_level': logging_level,
+                'worker_class': self.worker_class,
+                'job_class': self.job_class,
+                'serializer': self.serializer,
+            },
+            name=f'Worker {name} (WorkerPool {self.name})',
+        )
+
+
+def run_django_worker(*args: Any, **kwargs: Any) -> None:
+    # multiprocessing library default process start method may be
+    # `spawn` or `fork` depending on the host OS
+    if get_start_method() == 'spawn':
+        django.setup()
+
+    run_worker(*args, **kwargs)
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,6 +14,17 @@ jobs:
   build:
     runs-on: ubuntu-latest
     name: Python${{ matrix.python-version }}/Django${{ matrix.django-version }}
+
+    services:
+      postgres:
+        image: postgres:15
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: django_rq_test_db
+        ports:
+          - 5432:5432
+
     strategy:
       matrix:
         python-version: ["3.10", "3.11", "3.12"]
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,4 +49,4 @@ jobs:
 
     - name: Run Test
       run: |
-        `which django-admin` test django_rq --settings=django_rq.tests.settings --pythonpath=.
+        `which django-admin` test django_rq --settings=django_rq.tests.settings --pythonpath=. --noinput
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,7 +45,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install django==${{ matrix.django-version }}
-        pip install redis django-redis rq sentry-sdk rq-scheduler
+        pip install redis django-redis rq sentry-sdk rq-scheduler psycopg2-binary
 
     - name: Run Test
       run: |
--- a/django_rq/management/commands/rqworker-pool.py
+++ b/django_rq/management/commands/rqworker-pool.py
@@ -1,3 +1,4 @@
+import multiprocessing as mp
 import os
 import sys
 
@@ -7,7 +8,7 @@ from rq.serializers import resolve_serializer
 from django.core.management.base import BaseCommand
 
 from ...jobs import get_job_class
-from ...utils import configure_sentry
+from ...utils import configure_sentry, reset_db_connections
 from ...queues import get_queues
 from ...workers import get_worker_class
 from ...worker_pool import DjangoWorkerPool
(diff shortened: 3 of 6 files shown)


django_rq/worker_pool.py
```diff
@@ -0,0 +1,38 @@
...
+                'logging_level': logging_level,
+                'worker_class': self.worker_class,
+                'job_class': self.job_class,
+                'serializer': self.serializer,
+            },
+            name=f'Worker {name} (WorkerPool {self.name})',
+        )
+
+
+def run_django_worker(*args: Any, **kwargs: Any) -> None:
```
jackkinsella, line 32: FWIW The version of this function we're using in production was. 

```python
def run_django_worker(*args: Any, **kwargs: Any) -> None:
    alias = "default"

    try:
        connections[alias].close()
    except Exception:
        logger.error(f"Worker could not get a connection to the database with alias {alias}")

    run_worker(*args, **kwargs)

```

The exception catching and logging was transient and didn't stop us from using a 16 worker queue to process a few 100k events per day

jackkinsella: The `django.setup()` solution is likely better (but I don't understand it so can't offer feedback)

jackkinsella: I just checked production (Heroku - unix) and `get_start_method()` was "fork". I know that simply executing `run_worker()` will not work in that environment due to the psycog issues with the DB connection from pre-fork.

(Maybe this explains the failing CI test?)

django_rq/worker_pool.py
```diff
@@ -0,0 +1,38 @@
...
+    # multiprocessing library default process start method may be
+    # `spawn` or `fork` depending on the host OS
+    if get_start_method() == 'spawn':
+        django.setup()
+
+    run_worker(*args, **kwargs)
```
selwin, line 38: Similar to @jackkinsella's comment, we need to use our own `run_worker` command that calls `reset_db_connections` before running the original `run_worker` command: https://github.com/rq/django-rq/blob/025a53f379e48a9d5483421260d8052f143169bf/django_rq/utils.py#L150

django_rq/worker_pool.py
chromium7: Yes, I just realized that db connections are also closed pre-fork in the `rqworker` management command.
https://github.com/rq/django-rq/blob/d09421d48cdac38b9633f43c34a73cc906bc11ac/django_rq/management/commands/rqworker.py#L95-L96

I called `django.setup()` because the spawned process doesn't load the installed apps at all. This works fine when I manually tested it, but fails in unit test. Probably its because `django.setup()` is not setting up the project in testing environment which leads to the process setting up its own DB instance.

I'm currently using this django code for parallel test runner as a reference. I'll try to update the PR in a day or two.
https://github.com/django/django/blob/53719d6b5b745dd99b1ab9315afb242f706ebbf1/django/test/runner.py#L424-L432

PR #656, opened 2024-03-25
Add type annotations to most public functions

SpecLad:
Except for `django_rq.job`, which is too complicated to type properly.


selwin:
Thanks!

PR #659, opened 2024-04-13
get_worker: replace special kwargs with separate parameters

SpecLad:
This communicates more information to IDEs and type checkers, making the function easier to use.

Having regular parameters after `*args` has been possible since Python 3.0.


selwin:
Thanks!

PR #660, opened 2024-04-13
Make `rqstats -y` fail when PyYAML is not installed

SpecLad:
Currently, when PyYAML is not installed, `rqstats -y` prints an error message, but it still exits with a 0 exit code, indicating success. Fix that, along with a few related issues:

* The error message was printed to stdout rather than stderr.

* The name of the library in the message was incorrect (LibYAML is a C library; the Python library is PyYAML).

After this change, you need to install PyYAML to run the testsuite, so modify the GitHub workflow accordingly. Also, make sure to install all packages at once, to make sure that later install commands don't overwrite the result of the earlier ones.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,8 +33,8 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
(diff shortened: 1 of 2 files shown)


selwin:
Thanks!

PR #668, opened 2024-08-29
test type annotations with mypy

terencehonles:
I was going to contribute my typing for the `@job` decorator that was missing from https://github.com/rq/django-rq/pull/656, but I then realized there was no way to check if the annotations were correct, so I added `mypy` to the testing matrix

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,3 +39,24 @@ jobs:
     - name: Run Test
       run: |
         `which django-admin` test django_rq --settings=django_rq.tests.settings --pythonpath=.
+
+  mypy:
+    runs-on: ubuntu-latest
+    name: Type check
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v4.2.0
+      with:
+        python-version: "3.8"
+
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install django-stubs[compatible-mypy] rq types-redis
+
+    - name: Run Test
+      run: |
+        mypy django_rq
--- a/django_rq/decorators.py
+++ b/django_rq/decorators.py
@@ -1,8 +1,12 @@
 from rq.decorators import job as _rq_job
+from typing import TYPE_CHECKING, Union
 
 from django.conf import settings
 
 from .queues import get_queue
+
+if TYPE_CHECKING:
+    from rq import Queue
 
 
 def job(func_or_queue, connection=None, *args, **kwargs):
@@ -18,7 +22,7 @@ def job(func_or_queue, connection=None, *args, **kwargs):
     """
     if callable(func_or_queue):
         func = func_or_queue
-        queue = 'default'
+        queue: Union['Queue', str] = 'default'
     else:
         func = None
         queue = func_or_queue
@@ -30,13 +34,17 @@ def job(func_or_queue, connection=None, *args, **kwargs):
                 connection = queue.connection
         except KeyError:
             pass
+    else:
+        if connection is None:
+            connection = queue.connection
 
     RQ = getattr(settings, 'RQ', {})
     default_result_ttl = RQ.get('DEFAULT_RESULT_TTL')
     if default_result_ttl is not None:
         kwargs.setdefault('result_ttl', default_result_ttl)
 
-    decorator = _rq_job(queue, connection=connection, *args, **kwargs)
+    kwargs['connection'] = connection
+    decorator = _rq_job(queue, *args, **kwargs)
     if func:
         return decorator(func)
     return decorator
--- a/django_rq/management/commands/rqenqueue.py
+++ b/django_rq/management/commands/rqenqueue.py
@@ -22,9 +22,7 @@ class Command(BaseCommand):
         Queues the function given with the first argument with the
         parameters given with the rest of the argument list.
         """
-        verbosity = int(options.get('verbosity', 1))
-        timeout = options.get('timeout')
-        queue = get_queue(options.get('queue'))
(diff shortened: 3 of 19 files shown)


django_rq/decorators.py
```diff
@@ -30,13 +34,17 @@ def job(func_or_queue, connection=None, *args, **kwargs):
                 connection = queue.connection
         except KeyError:
             pass
+    else:
+        if connection is None:
+            connection = queue.connection
```
terencehonles, line 39: I think this was an oversight in the previous code?

django_rq/decorators.py
```diff
@@ -30,13 +34,17 @@ def job(func_or_queue, connection=None, *args, **kwargs):
...
 
     RQ = getattr(settings, 'RQ', {})
     default_result_ttl = RQ.get('DEFAULT_RESULT_TTL')
     if default_result_ttl is not None:
         kwargs.setdefault('result_ttl', default_result_ttl)
 
-    decorator = _rq_job(queue, connection=connection, *args, **kwargs)
+    kwargs['connection'] = connection
```
terencehonles, line 46: This is just for mypy (https://github.com/python/mypy/issues/6799) and rather than adding a type ignore I figured I'd update the kwargs, since it would never contain the connection based on the function's definition

django_rq/management/commands/rqenqueue.py
```diff
@@ -22,9 +22,7 @@ def handle(self, *args, **options):
         Queues the function given with the first argument with the
         parameters given with the rest of the argument list.
         """
-        verbosity = int(options.get('verbosity', 1))
```
terencehonles, line 25: These will all exist, and using indexed access will tell mypy they are not `None`

.github/workflows/test.yml
```diff
@@ -39,3 +39,24 @@ jobs:
...
+    runs-on: ubuntu-latest
+    name: Type check
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v4.2.0
+      with:
+        python-version: "3.8"
```
terencehonles, line 53: Picking an old supported Python version since it doesn't it's tested otherwise, and this will make sure the type annotations don't break the older Pythons

django_rq/py.typed
terencehonles, line 1: Export type annotations

django_rq/tests/settings.py
```diff
@@ -20,13 +20,6 @@
     except ImportError:
         REDIS_CACHE_TYPE = 'none'
 
-try:
-    from django.utils.log import NullHandler
```
terencehonles, line 24: This was removed from Django way before any of the supported versions

django_rq/tests/test_views.py
```diff
@@ -407,7 +409,7 @@ def test_action_stop_jobs(self):
         self.assertEqual(len(canceled_job_registry), len(job_ids))
 
         for job_id in job_ids:
-            self.assertIn(job_id, canceled_job_registry)
+            self.assertIn(job_id, canceled_job_registry)  # type: ignore[arg-type]
```
terencehonles, line 412: bug in `registry.__contains__` that breaks `assertIn`

django_rq/tests/tests.py
```diff
@@ -722,7 +718,7 @@ def test_commandline_verbosity_affects_logging_level(self, setup_loghandlers_moc
             setup_loghandlers_mock.assert_called_once_with(expected_level[verbosity])
 
     @override_settings(RQ={'SCHEDULER_CLASS': 'django_rq.tests.fixtures.DummyScheduler'})
-    def test_scheduler_default_timeout(self):
+    def test_scheduler_default(self):
```
terencehonles, line 721: This test was overridden by the one that comes after it (thanks mypy!)

django_rq/utils.py
```diff
@@ -116,12 +118,26 @@ def get_scheduler_statistics():
...
+        ]
+    ] = None,
+) -> List[Job]:
     """Fetch jobs in bulk from Redis.
     1. If job data is not present in Redis, discard the result
     2. If `registry` argument is supplied, delete empty jobs from registry
     """
-    jobs = Job.fetch_many(job_ids, connection=queue.connection, serializer=queue.serializer)
+    jobs = cast(
+        List[Optional[Job]], Job.fetch_many(job_ids, connection=queue.connection, serializer=queue.serializer)
```
terencehonles, line 139: this is a bug in `fetch_many` and mypy will complain that `job is None` can't happen otherwise

django_rq/views.py
```diff
@@ -403,7 +404,12 @@ def clear_queue(request, queue_index):
             queue.empty()
             messages.info(request, 'You have successfully cleared the queue %s' % queue.name)
         except ResponseError as e:
-            if 'EVALSHA' in e.message:
+            try:
+                suppress = 'EVALSHA' in e.message  # type: ignore[attr-defined]
```
terencehonles, line 408: I can't find this ever existing, but it seems to have worked for the PR author?

setup.cfg
```diff
@@ -1,2 +1,24 @@
 [bdist_wheel]
-universal = 1
\ No newline at end of file
+universal = 1
+
+[mypy]
+allow_redefinition = true
+check_untyped_defs = true
```
terencehonles, line 6: treat untyped code as `Any` to make sure everything is checked even if it ends up not being very useful

setup.cfg
```diff
@@ -1,2 +1,24 @@
...
+pretty = true
+show_error_codes = true
+show_error_context = true
+warn_redundant_casts = true
+warn_unused_ignores = true
+warn_unreachable = true
```
terencehonles, line 12: warn in case mypy thinks this case can never happen (in case it actually can and the type checking is not checking a path that should be checked)

selwin:
Thanks!

PR #669, opened 2024-08-29
Type `@job` decorator

terencehonles:
This change adds typing to the `@job` decorator which was missing from https://github.com/rq/django-rq/pull/656.


selwin:
Thanks!

PR #670, opened 2024-09-01
Fix broken /admin/background-tasks web UI when using Sentinel

### LINKED ISSUE #642: Sentinel support broken since 2.9.0
Hello altogether,

while using redis sentinel we've run into the following exception with django-rq 2.9.0 and 2.10.1:

```python
  File "/opt/netbox/venv/lib/python3.11/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/venv/lib/python3.11/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/venv/lib/python3.11/site-packages/django/views/decorators/cache.py", line 62, in _wrapper_view_func
    response = view_func(request, *args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/venv/lib/python3.11/site-packages/django/contrib/auth/decorators.py", line 23, in _wrapper_view
    return view_func(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/venv/lib/python3.11/site-packages/django_rq/views.py", line 37, in stats
    **get_scheduler_statistics(),
      ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/netbox/venv/lib/python3.11/site-packages/django_rq/utils.py", line 106, in get_scheduler_statistics
    conn_key = f"{connection_kwargs['host']}:{connection_kwargs['port']}/{connection_kwargs.get('db', 0)}"
                  ^^^^^^^^^^^^^^^^^^^^^^^^^

Exception Type: KeyError at /admin/background-tasks/
Exception Value: 'host'
```

Sentinel configuration does not have a `host` key, hence the `KeyError`.

The cause can be found here: https://github.com/rq/django-rq/blob/v2.10.1/django_rq/utils.py#L106

It was introduced with this commit: https://github.com/rq/django-rq/commit/df3271341a3668717844dcdc9a4a77e64a6316d0

I like that tests were added to test this new functionality as well, but they should also cover sentinel setups, as this library supports them and even provides example configuration as simple in the README.md for it.

We've stumbled over this problem via usage of NetBox, a Django based application leveraging django-rq: https://github.com/netbox-community/netbox/issues/14752

NeodymiumFerBore:
An attempt to fix #642.

## Bug recap

When using Sentinel, in `get_scheduler_statistics` function,  `queue.connection.connection_pool.connection_kwargs` does not have a `host` key, making the function to raise a `KeyError` exception when it tries to define `conn_key` based on `connection['host']`. This makes the `/admin/background-tasks` page to return a 500 error, making it unusable.

## What does this fix do

This PR fixes this by building `conn_key` from all sentinels host:port pairs.

With this fix, a scheduler configured with Sentinel appears like this (with `mymaster` being the Sentinel Service, and `0` the configured db index for this scheduler):

```
Sentinel(sentinel-1:26379,sentinel-2:26379,sentinel-3:26379)://mymaster/0
```
django-rq-fixed-admin-bg-tasks-sentinel

If you don't agree with this syntax, please let me know and I change it.

## Alternative solution

An alternative solution would be to build the `conn_key` from the current Sentinel master for the given service, but I'm not sure what you prefer. It would make the code a bit easier to read, and would make the resulting string in the UI much shorter. But it's less clear about the config used for this scheduler.

[NeodymiumFerBore/django-rq#2](https://github.com/NeodymiumFerBore/django-rq/pull/2/files) implements this solution, if you want to take a look...

### DIFF
--- a/django_rq/utils.py
+++ b/django_rq/utils.py
@@ -103,7 +103,18 @@ def get_scheduler_statistics():
         # jobs in more than one of them
         queue = get_queue_by_index(index)
         connection = queue.connection.connection_pool.connection_kwargs
-        conn_key = f"{connection['host']}:{connection.get('port', 6379)}/{connection.get('db', 0)}"
+
+        # When using Sentinel, connection does not have a host key, so build conn_key from all sentinels host:port pairs.
+        if 'connection_pool' in connection and hasattr(connection['connection_pool'], 'sentinel_manager'):
+            pool = connection['connection_pool']
+            sentinels = ','.join([
+                f"{s.connection_pool.connection_kwargs.get('host')}:{s.connection_pool.connection_kwargs.get('port')}"
+                for s in pool.sentinel_manager.sentinels
+            ])
+            conn_key = f"Sentinel({sentinels})://{pool.service_name}/{pool.sentinel_manager.connection_kwargs.get('db', 0)}"
+        else:
+            conn_key = f"{connection['host']}:{connection.get('port', 6379)}/{connection.get('db', 0)}"
+
         if conn_key not in schedulers:
             try:
                 scheduler = get_scheduler(config['name'])


moonrail:
Tested in context of NetBox 4.1.3 and am able to confirm that this fixes issue #642 and adds back Sentinel support.

Thank you @NeodymiumFerBore :+1:

NeodymiumFerBore:
@moonrail PR #671 got merged and fixes the same issue, with cleaner code. Can you test again with #671 ?
This PR should get closed shortly. If you confirm that #671 does the job, I'll close it myself.
Thank you for your feedback!

Issue #329, opened 2019-03-21
Allow setting DEFAULT_RESULT_TTL in RQ_QUEUES


rvernica:
`RQ_QUEUES` allows for setting `DEFAULT_TIMEOUT` when defining a queue in `settings.py`. Also, the `@job` decorator allows for setting `DEFAULT_RESULT_TTL` in `RQ` in `settings.py`.

It would be nice to allow setting `DEFAULT_RESULT_TTL` for queues in `RQ_QUEUES` in `settings.py`.

selwin:
Please open a PR for this :)

Dganzh:
How to pull request? When I push:

 `remote: Permission to rq/django-rq.git denied to Dganzh.`

selwin:
This might help: https://help.github.com/en/articles/creating-a-pull-request

:)

PR #693, opened 2025-02-05
Allow setting DEFAULT_RESULT_TTL in RQ_QUEUES

Sir-heed:
DEFAULT_RESULT_TTL is only set in the settings RQ, this value though does not apply to queue outside the `default` queue, other queues have their default value as 500 (this is default set in RQ)

This PR allows adding DEFAULT_RESULT_TTL to the RQ_QUEUES, hence controlling the time it takes for job result to leave the queue.

This is a response to the issue - #329 .

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -53,6 +53,7 @@ Installation
             'USERNAME': 'some-user',
             'PASSWORD': 'some-password',
             'DEFAULT_TIMEOUT': 360,
+            'DEFAULT_RESULT_TTL': 800,
             'REDIS_CLIENT_KWARGS': {    # Eventual additional Redis connection arguments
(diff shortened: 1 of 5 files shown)


selwin:
Thanks!

perry:
@Sir-heed @selwin this appears to have caused `result_ttl` from being ignored when set in the `@job` decorator and the RQ.DEFAULT_RESULT_TTL` setting.

Sir-heed:
Thanks for the notification @perry. I'll check it out.

Sir-heed:
@perry a fix has been raised #714

perez31000:
Hello, when the release tag will be available ? :)

Issue #697, opened 2025-04-04
ValueError: not enough values to unpack (expected 2, got 1)


Stephane-Ag:
I keep getting this error when trying to view active jobs when there's at least 1 active job.

It's being caused by this [function](http://github.com/rq/django-rq/blame/master/django_rq/utils.py#L162C21-L162C36):
```
def get_executions(queue, composite_keys: List[str]) -> List[Execution]:
    """Fetch executions in bulk from Redis.
    1. If execution data is not present in Redis, discard the result
    """
    executions = []
    for key in composite_keys:
        job_id, id = key.split(':')
        try:
            executions.append(Execution.fetch(id=id, job_id=job_id, connection=queue.connection))
        except ValueError:
            pass
    return executions
```

Why is a colon expected to be in the job_id?

I tried adding one to bypass this error but [rq explicitly doesn't allow this](https://github.com/rq/rq/blob/master/rq/job.py#L740) either and errors out with: `ValueError: id must not contain ":"`

`get_executions` seems to have been added in django-rq [v3.0](https://github.com/rq/django-rq/commit/fbb916c72370269d8fdc47124a0cbb77690a8938#diff-86374d856b018c779d498e105c46bd21bdd07450d63a28fa8fd440a9c42272cf) for rq v2.0 but rq [v2.0.1](https://github.com/rq/rq/commit/aa5c58f2229eee95b71361286c8d10ede953c948) added the no-colon requirement.

How can I solve this?

Stephane-Ag:
It looks like rq changed how the job_ids are parsed for the `StartedJobRegistry` in [v2.2](https://github.com/rq/rq/blame/d9fb4fda3c6edf183f85d8456bab507bf3341148/rq/registry.py#L170).
Downgrading rq to v2.1.0 currently fixes this issue for me.

PR #700, opened 2025-04-19
fix ``registry.get_job_id`` not returning execution id anymore

terencehonles:
This change uses the new ``registry.get_job_and_execution_ids`` instead of the updated ``registry.get_job_id`` which no longer returns the execution id.

fixes: #697


selwin:
Thanks!

Issue #699, opened 2025-04-15
Py.typed not included in package


sk-:
In PR #668 the file `py.typed` was added to the repository, however it is currently not being included in the built wheels. The reason for this is because the package metadata in file `setup.py` is specifying the package as `rq` instead of `django_rq`.

PR #701, opened 2025-04-21
fix typo in setup.py package_data

terencehonles:
This change fixes a typo in setup.py package_data which prevents the type information from being exported.

fixes: https://github.com/rq/django-rq/issues/699


selwin:
Thanks!

Issue #698, opened 2025-04-09
django 5.2 exception


momenkittaneh:
hi,
I've upgraded my packages to LTS with django 5.2
but when i enter job order details this exception occured

![Image](https://github.com/user-attachments/assets/53e558a4-bb73-4dc9-8fc7-7bf7c079e825)

Can you help me how to fix this issue
and thank you

daufinsyd:
I encounter the same error when using django-rq 3

terencehonles:
This looks like it's at least partially related to https://github.com/django/django/pull/19389, and will likely be fixed in Django.

PR #702, opened 2025-04-21
fix Django admin object-tools breaking job detail view

terencehonles:
This change removes the object-tools from the job detail-view.

fixes: https://github.com/rq/django-rq/issues/698
x-ref: https://code.djangoproject.com/ticket/36331
x-ref: https://github.com/django/django/pull/19389


selwin:
@terencehonles can you check why the tests were failing?

PR #706, opened 2025-05-02
fix JSON stats view and tests with redis-py 6.0

terencehonles:
x-ref: https://github.com/redis/redis-py/pull/3614
x-ref: https://github.com/redis/redis-py/pull/3622

### DIFF
--- a/django_rq/tests/utils.py
+++ b/django_rq/tests/utils.py
@@ -5,15 +5,18 @@ def get_queue_index(name='default'):
     """
     Returns the position of Queue for the named queue in QUEUES_LIST
     """
-    queue_index = None
     connection = get_connection(name)
     connection_kwargs = connection.connection_pool.connection_kwargs
+
     for i in range(0, 100):
         try:
             q = get_queue_by_index(i)
         except AttributeError:
             continue
-        if q.name == name and q.connection.connection_pool.connection_kwargs == connection_kwargs:
-            queue_index = i
-            break
-    return queue_index
+        if q.name == name:
+            # assert that the connection is correct
+            assert q.connection.connection_pool.connection_kwargs == connection_kwargs
+
+            return i
+
+    return None
--- a/django_rq/utils.py
+++ b/django_rq/utils.py
@@ -67,9 +67,10 @@ def get_statistics(run_maintenance_tasks=False):
         else:
             oldest_job_timestamp = "-"
 
-        # parse_class and connection_pool are not needed and not JSON serializable
-        connection_kwargs.pop('parser_class', None)
+        # remove unneeded properties which are not serializable in JSON
         connection_kwargs.pop('connection_pool', None)
+        connection_kwargs.pop('parser_class', None)
+        connection_kwargs.pop('retry', None)
 
         queue_data = {
             'name': queue.name,


django_rq/utils.py
```diff
@@ -67,9 +67,10 @@ def get_statistics(run_maintenance_tasks=False):
         else:
             oldest_job_timestamp = "-"
 
-        # parse_class and connection_pool are not needed and not JSON serializable
-        connection_kwargs.pop('parser_class', None)
+        # remove unneeded properties which are not serializable in JSON
         connection_kwargs.pop('connection_pool', None)
+        connection_kwargs.pop('parser_class', None)
+        connection_kwargs.pop('retry', None)
```
terencehonles, line 73: Do we want to try to serialize this?

django_rq/tests/utils.py
```diff
@@ -5,15 +5,36 @@ def get_queue_index(name='default'):
...
     """
-    queue_index = None
     connection = get_connection(name)
-    connection_kwargs = connection.connection_pool.connection_kwargs
+    connection_kwargs = dict(connection.connection_pool.connection_kwargs)
+
+    special_case_retry = False
+    if (retry := connection_kwargs.get('retry')) and retry.__class__.__eq__ == object.__eq__:
+        special_case_retry = True
+        del connection_kwargs['retry']
```
terencehonles, line 14: Special casing can be removed if https://github.com/redis/redis-py/pull/3628 is accepted and released in a minor / point release and we don't need to support all previous releases (or the support for 6.0 is dropped, which will probably be in a long time)

selwin: It seems like your PR was merged yesterday. https://github.com/redis/redis-py/pull/3628 . Should we just wait for redis-py 6.0.1 to be released before merging this in?

terencehonles: Up to you, not sure how long that will be

selwin: I think I might just exclude redis-py 6.0 from dependency. Will cut a release soon.

terencehonles: I'll update this to not have the workaround then. I'm assuming you will have something like `redis-py!=6.0.0` in the dependency specifiers, so that when 6.0.1 is released this will work properly still.

selwin: Yes, I merged this in just a few minutes ago: https://github.com/rq/rq/pull/2244

django_rq/tests/utils.py
```diff
@@ -5,15 +5,18 @@ def get_queue_index(name='default'):
...
     for i in range(0, 100):
         try:
             q = get_queue_by_index(i)
         except AttributeError:
             continue
-        if q.name == name and q.connection.connection_pool.connection_kwargs == connection_kwargs:
-            queue_index = i
-            break
-    return queue_index
+        if q.name == name:
```
terencehonles, line 16: I updated this to only check the name, since this method was skipping the queue (and breaking on an `IndexError` if the connection wasn't the "same")

django_rq/tests/utils.py
terencehonles: Ok, I updated this PR for you

selwin:
Thanks!

PR #707, opened 2025-05-04
fix pyright check

alphatownsman:
based on [pyright doc](https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#library-interface), imported symbols are considered private by default. 

without this change, pyright will report error: `"get_queue" is not exported from module "django_rq" (reportPrivateImportUsage)`


selwin:
Thanks!

PR #715, opened 2025-07-23
Update packaging to use modern packaging with a pyproject.toml

terencehonles:
This change updates and modernizes the packaging to use hatchling as that is used in the main RQ project.

This fixes deprecation warnings that pip is starting to produce:

> DEPRECATION: Building 'django-rq' using the legacy setup.py bdist_wheel mechanism, which will be removed in a future version. pip 25.3 will enforce this behaviour change. A possible replacement is to use the standardized build interface by setting the `--use-pep517` option, (possibly combined with `--no-build-isolation`), or adding a `pyproject.toml` file to the source tree of 'django-rq'. Discussion can be found at https://github.com/pypa/pip/issues/6334

fixes: https://github.com/rq/django-rq/issues/553


terencehonles:
I built the wheels and sources before and after this change and they look good.

terencehonles:
> Hey @terencehonles thanks for this. Just wanted to check whether this will also fix [#716 (comment)](https://github.com/rq/django-rq/issues/716#issuecomment-3124277993)

It looks like #716 is an issue of when the release was cut. Looking at [setup.py at the v3.0.1 tag](https://github.com/rq/django-rq/blob/v3.0.1/setup.py) it does indeed not provide the extra, and looking at the commit history https://github.com/rq/django-rq/commit/4ccb4127cb0db291cd489c3a74a534c2d967483b is only on `master` and not part of any release.

So this provides the same extras as that commit does, but really just a new release will fix the issue reported in #716

PR #718, opened 2025-07-27
Change url token to header auth and add to /metrics/ aswell

RasmusThing:
This PR adds token capability to the /metrics/ endpoint, just like how it's done on the /stats.json/ endpoint.

If you need to access the metrics view via other HTTP clients (for monitoring purposes), you can define RQ_API_TOKEN and access it via /django-rq/metrics/<API_TOKEN>.

-----

**Update:**
This PR replaces the token URL and replaces it with header authorization by using the bearer format.
Additionally, it adds the same token auth capability to the new prometheus /metrics/ endpoint and adds a link to the metrics page alongside the json link in the admin panel.

### DIFF
--- a/django_rq/stats_views.py
+++ b/django_rq/stats_views.py
@@ -1,3 +1,5 @@
+from secrets import compare_digest
+
 from django.contrib import admin
 from django.contrib.admin.views.decorators import staff_member_required
 from django.http import Http404, HttpResponse, JsonResponse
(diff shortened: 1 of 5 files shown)


selwin:
> I agree. I just copied the mechanics from the token auth done on the stats.json endpoint.
And yes, it would be wiser not to have the API_TOKEN exposed in the URL for both endpoints.

Do most scrapers support token authentication in headers? If possible I'd like the `/metrics/` endpoint to only support header based auth and `/stats.json` endpoints to support both methods of auth. 

I can then drop the URL based token auth from the `stats.json` endpoint when I release Django-RQ 4.0.

django_rq/stats_views.py
```diff
@@ -47,10 +64,12 @@ def stats(request):
     return render(request, 'django_rq/stats.html', context_data)
 
 
-def stats_json(request, token=None):
-    if request.user.is_staff or (token and token == API_TOKEN):
-        return JsonResponse(get_statistics())
+@never_cache
+def stats_json(request):
+    if not is_authorized(request):
+        return JsonResponse(
```
selwin, line 70: @RasmusThing I'm not ready to break backward compatibility with the current `stats_json` view yet, I'd like to give users a warning and time to transition.

Do you mind creating a new view which maps to the url `/stats.json` with header token auth? This way we don't need to modify the current `def stats_json(request, token=None)` view.

selwin:
Thanks!

PR #721, opened 2025-08-02
Move "View Metrics" inside existing <p>

RasmusThing:
Just a small PR to make the two "View as JSON" and "View Metrics" link to be inline instead of seperate rows.


selwin:
Thanks!

PR #752, opened 2025-12-07
RQ dashboard

selwin:
## Summary by CodeRabbit

* **New Features**
  * Added a standalone rq-dashboard CLI to run the RQ Dashboard as an independent app with automatic minimal Django setup, persistent secret-key handling, optional static collection, and built-in admin + dashboard routing.
  * Added a console entrypoint providing the new rq-dashboard command.

* **Tests**
  * Added tests for CLI config loading, secret-key persistence, CLI argument parsing, and dashboard/admin URL routing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rq/django-rq/pull/752)


django_rq/dashboard/cli.py
```diff
@@ -0,0 +1,359 @@
+#!/usr/bin/env python
```
terencehonles: Since you've added the [script entrypoint in your pyproject.toml](https://github.com/rq/django-rq/pull/752/changes#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R45) I'm not sure it makes sense to also make this an executable


```suggestion
```

(also check the +x bit)

django_rq/dashboard/cli.py
```diff
@@ -0,0 +1,359 @@
...
+# Optional settings — uncomment to use:
+# DEBUG = True
+# ALLOWED_HOSTS = ["dashboard.example.com"]  # default: ["127.0.0.1", "localhost"]
+'''
+
+
+def load_config(config_path: str) -> dict[str, Any]:
+    """Load RQ configuration from a Python file."""
+    config_path = str(Path(config_path).resolve())
+    if not Path(config_path).exists():
```
terencehonles: [`importlib` supports path-like objects](https://docs.python.org/3/library/importlib.html#importlib.util.spec_from_file_location)

```suggestion
    config_path = Path(config_path).resolve()
    if not config_path.exists():
```

django_rq/dashboard/cli.py
```diff
@@ -0,0 +1,359 @@
...
+        print(f"Error: Config file not found: {config_path}")
+        sys.exit(1)
```
terencehonles: Errors should be written to stderr, and sys.exit will do this for you:


```suggestion
        sys.exit(f"Error: Config file not found: {config_path}")
```

django_rq/dashboard/cli.py
```diff
@@ -0,0 +1,359 @@
...
+
+    spec = importlib.util.spec_from_file_location("rq_config", config_path)
+    if spec is None or spec.loader is None:
+        print(f"Error: Could not load config file: {config_path}")
+        sys.exit(1)
```
terencehonles: Same as above

```suggestion
        sys.exit(f"Error: Could not load config file: {config_path}")
```

django_rq/dashboard/cli.py
```diff
@@ -0,0 +1,359 @@
...
+
+    config_module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(config_module)
+
+    if not hasattr(config_module, 'RQ_QUEUES'):
+        print("Error: Config file must define RQ_QUEUES")
+        sys.exit(1)
```
terencehonles:
```suggestion
        sys.exit("Error: Config file must define RQ_QUEUES")
```

django_rq/dashboard/cli.py
```diff
@@ -0,0 +1,359 @@
...
+
+    if not hasattr(config_module, 'SECRET_KEY'):
+        print(
+            "Error: Config file must define SECRET_KEY. "
+            "Run `rq-dashboard init` to generate a fresh config with a random key."
+        )
+        sys.exit(1)
```
terencehonles:
```suggestion
        sys.exit(
            "Error: Config file must define SECRET_KEY. "
            "Run `rq-dashboard init` to generate a fresh config with a random key."
        )
```

django_rq/dashboard/cli.py
```diff
@@ -0,0 +1,359 @@
...
+def resolve_config_path(explicit: Optional[str]) -> Path:
+    """Resolve which config file to load."""
+    if explicit:
+        return Path(explicit)
+
+    cwd_config = Path.cwd() / SAMPLE_CONFIG_FILENAME
+    if cwd_config.exists():
+        return cwd_config
+
+    print("rq-dashboard requires a config file.")
```
terencehonles: This should also probably go to stderr...

django_rq/dashboard/cli.py
```diff
@@ -0,0 +1,359 @@
...
+    print("    rq-dashboard init")
+    sys.exit(1)
+
+
+def write_sample_config() -> None:
+    """Write a starter rq_dashboard_config.py into the current directory."""
+    path = Path.cwd() / SAMPLE_CONFIG_FILENAME
+    if path.exists():
+        print(f"{path} already exists. Refusing to overwrite.")
+        sys.exit(1)
```
terencehonles:
```suggestion
        sys.exit(f"{path} already exists. Refusing to overwrite.")
```

django_rq/dashboard/cli.py
```diff
@@ -0,0 +1,359 @@
...
+    # },
+}
+
+# Optional settings — uncomment to use:
+# DEBUG = True
+# ALLOWED_HOSTS = ["dashboard.example.com"]  # default: ["127.0.0.1", "localhost"]
+'''
+
+
+def load_config(config_path: str) -> dict[str, Any]:
```
terencehonles: It looks like you can accept a Path here

```suggestion
def load_config(config_path: str | Path) -> dict[str, Any]:
```

django_rq/dashboard/cli.py
```diff
@@ -0,0 +1,359 @@
...
+        build_parser().print_help()
+        sys.exit(0)
+
+    if args.command == 'init':
+        write_sample_config()
+        return
+
+    if args.command in ('createsuperuser', 'changepassword'):
+        config_path = resolve_config_path(args.config)
+        config = load_config(str(config_path))
```
terencehonles: You can just change the signature and this is fine:

```suggestion
        config = load_config(config_path)
```

django_rq/dashboard/cli.py
```diff
@@ -0,0 +1,359 @@
...
+        configure_django(config, config_path)
+        # Ensure auth tables exist before the auth command runs.
+        call_command('migrate', verbosity=0)
```
terencehonles: This is the same in all the branches, so you should just move it up and out of the `if`

terencehonles (approved):
I didn't boot this up, but seems like it should work fine, and it has the benefit of kicking the tires on `django_rq`'s Django integration so that's a plus.

Since this is just using the dev-server, I think it should be made clear that this should not be used on the public internet. Otherwise I think it makes sense to bundle this with some other server i.e. uvicorn.

This is possibly taking things too far: looking at Django's source code, and how you're instantiating the configuration, I think you can move this to be a single module by:

- moving `cli.py` to `django_rq/dashboard.py` you can add `__all__ = ['main']` to only export that
- move your `urlpatterns` and imports from `urls.py` to `django_rq/dashboard.py`
- using the setting `ROOT_URLCONF = 'django_rq.dashboard'`  will have it point back to the same module.

PR #754, opened 2025-12-16
fix Django admin integration including preserving compatibility

terencehonles:
This change fixes the Django admin integration added in https://github.com/rq/django-rq/pull/747. As is, the integration adds the views, but they do not reverse properly without also including the URLs via `include("django_rq.urls")`. When using Django's `ModelAdmin`, the URLs registered will be in the "admin" namespace, but the templates were not updated to include a namespace so they fail to resolve.

This change conditionally resolves with either the prefix `django_rq:` or `admin:django_rq_` depending on how the view is called (via the admin app or not) and the URLs are also registered with the corresponding prefix.

In order to properly test this integration, the default URL conf has been updated to not include the django_rq URLS via `include`, and a separate `default_with_custom_mount_urls` URL conf is provided in order to test integrations using both `include` and the admin integration.

### DIFF
--- a/tests/test_admin_integration.py
+++ b/tests/test_admin_integration.py
@@ -49,21 +49,21 @@ class AuthenticatedAdminURLTest(TestCase):
 
     def test_stats_json_url(self):
         """Verify stats JSON endpoint accessible via admin URLs"""
-        response = self.client.get(reverse('rq_home_json'))
+        response = self.client.get(reverse('admin:django_rq_home_json'))
         self.assertEqual(response.status_code, 200)
         self.assertEqual(response['Content-Type'], 'application/json')
 
     def test_queue_jobs_url(self):
         """Verify queue jobs view accessible via admin URLs"""
         queue_index = get_queue_index('default')
-        response = self.client.get(reverse('rq_jobs', args=[queue_index]))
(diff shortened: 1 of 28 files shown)


tests/test_admin_integration.py
```diff
@@ -72,39 +72,38 @@ def test_job_detail_url(self):
...
         """Verify reverse() with URL names still resolves correctly"""
-        # Test that URL names can be resolved
-        # Note: When accessed via admin, URLs may need admin namespace
-        try:
-            url = reverse('rq_home')
-            self.assertTrue(url.startswith('/'))
-        except Exception:
-            # If standard reverse fails, try with admin namespace
-            url = reverse('admin:rq_home')
-            self.assertTrue(url.startswith('/admin/'))
```
terencehonles: This test case was showing that there was a problem

tests/test_views.py
terencehonles: I didn't test both paths for this view, but it could be updated if that was important

terencehonles:
I'll try to take a look if I have a chance. I had noticed the PR when looking into the issue for this PR and seeing what had changed recently and if anyone else had noticed the issue. I will be taking off for holidays and then an extended break, so may not get to it, but we'll see. Hope you have a nice holiday period yourself.

PR #788, opened 2026-04-27
Dropped support for Python 3.9.

p-r-a-v-i-n:
see https://www.djangoproject.com/download/

I think we should drop support for django 4.2, django 5.0 and django 5.1 bcs they reached to their EOL  and by virtue of it we need to drop python 3.9.

## Summary by CodeRabbit

* **Chores**
  * Updated minimum Python version requirement from 3.9 to 3.10. Support for Python 3.8 and 3.9 has been discontinued.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,7 @@ readme = "README.md"
 license = "MIT"
 license-files = ["LICENSE.txt"]
 authors = [{ name = "Selwin Ong", email = "selwin.ong@gmail.com" }]
-requires-python = ">=3.9"
+requires-python = ">=3.10"
 dependencies = ["django>=4.2", "rq>=2.6.1", "redis>=3.5"]
 classifiers = [
     "Development Status :: 5 - Production/Stable",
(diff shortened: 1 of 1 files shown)


selwin:
I don't usually drop support unless there are benefits (simplified code, new syntax etc). In this case, I'm ok with dropping Python 3.9 since Python 3.10 has several typing improvements, but not Django 4.2.

p-r-a-v-i-n:
So are you ok with dropping django 5.0 and django 5.1?

selwin:
> So are you ok with dropping django 5.0 and django 5.1?

What are the benefits of dropping support for Django 5.0 and 5.1?

p-r-a-v-i-n:
> What are the benefits of dropping support for Django 5.0 and 5.1?

will reduce some of  maintenance burden and I may be wrong but i don't see any benefits in keeping them.

selwin:
> will reduce some of maintenance burden and I may be wrong but i don't see any benefits in keeping them.

Will deprecate the next time we need to make changes to accommodate them :). The benefits is that people running older versions can still upgrade their RQ/Django-RQ versions should they need to.

selwin:
Found a reason to drop Python 3.9 :). I was working on this PR and having to use `Optional[BaseRegistry]` over `BaseRegistry | None` was indeed annoying https://github.com/rq/django-rq/pull/794/changes

PR #805, opened 2026-06-29
Return 404 for missing queue detail

### LINKED ISSUE #804: Queue detail page throws `IndexError` for non-existent queues
If you're using the Django-RQ dashboard and try to access a queue detail page by URL with the wrong queue index, the result is a 500 error response due to an `IndexError` bubbled up from `get_queue_by_index`.

Expected: this would return a 404 (and not trigger a bug event in the host system).

bennylope:
Fixes #804 

Adds a `try`/`except` guard to `get_queue_by_index`, which in turn requests an unvalidated indexed value from the queues.

Adds a simple test that is robust against additions to the number of configured test queues.

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid queue detail URLs now return a proper 404 error instead of failing unexpectedly.
  * Improved handling for missing queues when viewing queue details in the admin interface.

* **Tests**
  * Added coverage to verify the queue details page returns 404 for non-existent queues.

### DIFF
--- a/django_rq/views.py
+++ b/django_rq/views.py
@@ -82,7 +82,11 @@ def jobs(request: HttpRequest, queue_index: int) -> HttpResponse:
 @never_cache
 @staff_member_required
 def queue_details(request: HttpRequest, queue_index: int) -> HttpResponse:
-    queue = get_queue_by_index(queue_index)
+    try:
+        queue = get_queue_by_index(queue_index)
+    except IndexError:
+        raise Http404(f"Couldn't find queue with this ID: {queue_index}")
+
     connection = queue.connection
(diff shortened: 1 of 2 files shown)


selwin:
@bennylope mind fixing the failing type check so I can merge this in?

bennylope:
@selwin Sure! I originally left this alone since it's not caused by any changes in this branch. In the interests of atomicity I set that fix up as [a separate issue and PR](https://github.com/rq/django-rq/issues/807).

If you run `mypy` on the current master branch you'll see the same failure.

selwin:
Fixed here https://github.com/rq/django-rq/pull/806 @bennylope

selwin:
Thanks!

PR #811, opened 2026-07-09
Rename Django auth permission for Django RQ from "view" to "admin"

terencehonles:
This change renames the "view" permission to "admin" as the provided Django RQ admin views allow data modification and an admin may incorrectly grant unintended permissions if the codename is used in a query.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated dashboard access permissions so the admin page now uses a single, consistent permission label.
  * Existing access rights were migrated to preserve dashboard availability during the change.


terencehonles:
I tested the migration works appropriately in our app

selwin:
Thanks!

PR #823, opened 2026-09-10
Respect Django timezone in cron scheduler

### LINKED ISSUE #822: RQ cron scheduler does not respect Django TIME_ZONE
## Summary

Cron expressions registered through `django_rq` are evaluated in UTC instead of the timezone configured in Django's `TIME_ZONE`.

For example, with `TIME_ZONE = "Asia/Jakarta"`, a cron expression intended to run at `08:25` is scheduled for `08:25 UTC` and therefore runs at `15:25` Jakarta time.

## Environment

- django-rq: 4.2.0
- rq: 2.12.0
- Django `TIME_ZONE`: `Asia/Jakarta`
- Django `USE_TZ`: `True`

## Configuration

```python
# settings.py

TIME_ZONE = "Asia/Jakarta"
USE_TZ = True

# cron_config.py

from rq import cron

cron.register(
    func=some_job,
    queue_name="default",
    cron="25 8-16 * * *",
)
```

The scheduler is started through django-rq's rqcron management command.

## Expected behavior

The cron expression should be interpreted using Django's configured timezone.

With TIME_ZONE = "Asia/Jakarta":

25 8-16 * * *

should enqueue the job at:

08:25, 09:25, ..., 16:25 Asia/Jakarta

Alternatively, django-rq could provide an explicit setting or argument controlling the timezone used for cron evaluation.

## Actual behavior

The cron expression is interpreted in UTC. The first execution is calculated as 08:25 UTC, which django-rq's admin correctly displays as 15:25 Asia/Jakarta.

Example from the admin:

Schedule:     Cron: 25 8-16 * * *
Next enqueue: 2026-09-10 15:25:00

## Cause

RQ's CronJob initializes croniter using rq.utils.now():

```py
cron_iter = croniter(self.cron, now())

rq.utils.now() always returns UTC:

def now():
    return datetime.datetime.now(datetime.timezone.utc)
```

Because DjangoCronScheduler delegates cron registration to RQ's CronScheduler, Django's active timezone and settings.TIME_ZONE do not affect cron evaluation.

## Suggested behavior

Would django-rq consider supporting a configurable cron timezone, defaulting to Django's configured timezone?

Possible approaches might include:

1. A setting such as:

   RQ_CRON_TIMEZONE = TIME_ZONE

2. A timezone argument accepted by django_rq.cron.register().
3. Converting the base datetime to Django's configured timezone before passing it to croniter.

The calculated enqueue time can still be persisted as UTC; only the evaluation of the cron expression needs to happen in the selected timezone.

I understand that the underlying behavior originates in RQ's built-in CronScheduler. If this must be implemented upstream in RQ instead, clarification or documentation in django-rq would still be helpful. Currently, it is reasonable for Django users to expect cron expressions started through manage.py rqcron to respect Django's timezone configuration.

## Workaround

For a fixed-offset timezone such as Asia/Jakarta, the cron expressions can be manually converted to UTC:

25 8-16 * * *  ->  25 1-9 * * *

However, manual conversion is error-prone and is not sufficient for timezones that observe daylight-saving time.

chromium7:
## Summary

- evaluate cron expressions using Django's configured `TIME_ZONE`
- convert calculated enqueue times back to UTC for RQ storage and comparisons
- allow overriding cron evaluation with `RQ_CRON_TIMEZONE`
- require RQ 2.11+ and use its `next_enqueue_time` API
- import `croniter` directly while relying on RQ's dependency declaration

Fixes #822.

## Testing

- `uv run --with '.[testing]' pytest tests/test_cron.py -q` (18 passed)
- `ruff check django_rq/cron.py tests/test_cron.py`
- `git diff --check`

The full suite was also attempted locally under Python 3.14, but existing worker tests segfault in `setproctitle` after forking and the local RQ 2.8 environment has unrelated compatibility failures.

## Summary by CodeRabbit

- **New Features**
  - Cron schedules now consistently use Django’s configured timezone, including across daylight-saving transitions.
  - Cron job names now always link to their detail pages.
  - Job history is displayed whenever available.

- **Documentation**
  - Updated cron documentation to describe Django’s `TIME_ZONE` as the source for cron scheduling.

- **Changes**
  - Removed support for overriding cron timezone evaluation with `RQ_CRON_TIMEZONE`.

### DIFF
--- a/django_rq/cron.py
+++ b/django_rq/cron.py
@@ -3,16 +3,58 @@ from collections.abc import Sequence
 from datetime import datetime, timezone
 from functools import cached_property
 from typing import Any, Callable, Optional, cast
+from zoneinfo import ZoneInfo
 
+from django.conf import settings
+from django.utils.timezone import get_default_timezone
 from redis import Redis
-from rq.cron import CronJob, CronScheduler
-from rq.utils import as_text
+from rq.cron import CronJob, CronScheduler, croniter
+from rq.utils import as_text, now
 
 from .connection_utils import get_connection, get_redis_connection, get_unique_connection_configs
 from .settings import get_queues_map
 
 # `CronJob.get_job_ids()` and the job history sorted set backing it were added in RQ 2.11
 CRON_JOB_HISTORY_SUPPORTED = hasattr(CronJob, 'get_job_ids')
+
+
+def get_cron_timezone():
+    """Return the timezone used to evaluate cron expressions."""
+    configured_timezone = getattr(settings, 'RQ_CRON_TIMEZONE', None)
+    if configured_timezone is None:
+        return get_default_timezone()
+    if isinstance(configured_timezone, str):
+        return ZoneInfo(configured_timezone)
+    return configured_timezone
+
+
+class DjangoCronJob(CronJob):
+    """An RQ cron job whose cron expressions are evaluated in Django's timezone."""
+
+    def __init__(self, *args: Any, **kwargs: Any):
+        super().__init__(*args, **kwargs)
+        if self.cron:
+            next_time = self._get_next_cron_time(now())
+            if hasattr(self, 'next_enqueue_time'):
+                self.next_enqueue_time = next_time
+            else:  # RQ < 2.11
+                self.next_run_time = next_time
+
+    def _get_next_cron_time(self, base_time: datetime) -> datetime:
+        local_time = base_time.astimezone(get_cron_timezone())
+        return croniter(self.cron, local_time).get_next(datetime).astimezone(timezone.utc)
+
+    def get_next_enqueue_time(self) -> datetime:
+        """Calculate the next enqueue time, evaluating cron expressions in local time."""
+        if self.cron:
+            return self._get_next_cron_time(self.latest_enqueue_time or now())
+        return super().get_next_enqueue_time()
+
+    def get_next_run_time(self) -> datetime:
+        """RQ < 2.11 compatibility alias for timezone-aware cron evaluation."""
+        if self.cron:
+            return self._get_next_cron_time(getattr(self, 'latest_run_time', None) or now())
+        return super().get_next_run_time()  # type: ignore[misc]
 
 
 def get_cron_job_history(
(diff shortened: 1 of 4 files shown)


django_rq/cron.py
```diff
@@ -15,6 +18,45 @@
...
+
+class DjangoCronJob(CronJob):
+    """An RQ cron job whose cron expressions are evaluated in Django's timezone."""
+
+    def __init__(self, *args: Any, **kwargs: Any):
+        super().__init__(*args, **kwargs)
+        if self.cron:
+            next_time = self._get_next_cron_time(now())
+            if hasattr(self, 'next_enqueue_time'):
+                self.next_enqueue_time = next_time
```
selwin: We can bump the minimum dependency to RQ 2.11 to simplify things

django_rq/cron.py
```diff
@@ -3,10 +3,13 @@
...
 from functools import cached_property
 from typing import Any, Callable, Optional, cast
+from zoneinfo import ZoneInfo
 
+from django.conf import settings
+from django.utils.timezone import get_default_timezone
 from redis import Redis
-from rq.cron import CronJob, CronScheduler
-from rq.utils import as_text
+from rq.cron import CronJob, CronScheduler, croniter
```
selwin: Can you import directly from croniter and not from rq.cron?

selwin:
Can you simplify these?
CleanShot 2026-09-11 at 17 08 06@2x

selwin:
Codex also found this DST bug.

CleanShot 2026-09-11 at 17 08 43@2x

If needed, I don't mind specifying a minimum croniter version.

django_rq/cron.py
```diff
@@ -15,6 +19,34 @@
 CRON_JOB_HISTORY_SUPPORTED = hasattr(CronJob, 'get_job_ids')
 
 
+def get_cron_timezone():
+    """Return the timezone used to evaluate cron expressions."""
+    configured_timezone = getattr(settings, 'RQ_CRON_TIMEZONE', None)
```
selwin, line 24: We shouldn't need to have a `RQ_CRON_TIMEZONE` setting, just use Django's timezone settings. I plan on adding a feature in RQ to override timezones, so we should use that after it ships.

